PR #89, opened 2020-06-14
Store database in user profile directory

### LINKED ISSUE #88: Store database in user profile directory
**Environment**
 - OS: Any

**Describe the solution you'd like**
`ScannedFiles.db` should be in the user's home directory instead of together with the executable files. This way the program- and data files would be separated.

I guess these directories should be `C:\Users\<user>\AppData\Roaming\VDF` in Windows and `/home/<user>/.config/vdf` in Linux or something similar.

belidzs:
Fixes #88

Only tested in Windows but should be OS-agnostic.

### DIFF
--- a/DuplicateFinderEngine/DatabaseHelper.cs
+++ b/DuplicateFinderEngine/DatabaseHelper.cs
@@ -10,10 +10,11 @@ using ProtoBuf;
 
 namespace DuplicateFinderEngine {
 	public static class DatabaseHelper {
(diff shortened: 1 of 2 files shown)


0x90d:
Thanks but VDF should be portable. Thus all files should be within VDF directory only.

smares:
Forgive my ignorance, but is this really a common use case? The DB is tightly coupled to the scanned directories for which the user running the application needs access rights and neither the locations, nor the rights are portable. They are specific to that OS instance, thus having the DB in the user profile directory makes more sense to me.

PR #91, opened 2020-06-14
Fix unable to find ffmpeg in PATH when using Windows

### LINKED ISSUE #90: Application doesn't find ffmpeg.exe in PATH
**Environment**
 - OS: Windows

**Describe the bug**
Application doesn't find ffmpeg.exe in PATH, it has to be copied next to the executable.

**Expected behavior**
ffmpeg.exe should be automatically discovered when it's available in PATH

belidzs:
The automatically discovered `FfmpegPath` was `ffmpeg.exe.exe`.

This PR fixes #90

Refactor Utils.cs
* add comments
* remove static constructor
* implement FindFfmpegDirectory
* assume ffmpeg and ffprobe is in the same directory

### DIFF
--- a/DuplicateFinderEngine/Utils.cs
+++ b/DuplicateFinderEngine/Utils.cs
@@ -6,61 +6,54 @@ using System.Runtime.InteropServices;
 
 namespace DuplicateFinderEngine {
 	public static class Utils {
-		public const string FFprobeExecutableName = "ffprobe";
-		public const string FFmpegExecutableName = "ffmpeg";
-		public static string FfmpegPath { get; }
-		public static string FfprobePath { get; }
-
+		public static string FFprobeExecutableName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "ffprobe.exe" : "ffprobe";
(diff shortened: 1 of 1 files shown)


0x90d:
This PR does much more than fixing the issue in #90, especially it changes the way the path is cached into non-cached. Everytime the path gets accessed it calls Path.Combine, I won't merge this one the way it is. But I leave it open for you to modify it. Preferrably  only fix #90 first, then make a new PR for further changes.

Revan654:
> 
> 
> This PR does much more than fixing the issue in #90, especially it changes the way the path is cached into non-cached. Everytime the path gets accessed it calls Path.Combine, I won't merge this one the way it is. But I leave it open for you to modify it. Preferrably only fix #90 first, then make a new PR for further changes.

What needs to be done is cache the original path through property (get; set;) or call an extension method. Whoever calls the method first caches the supplied path. Then locks the file, so it can't be moved or altered.

Path.Combine actually doesn't use much resources(Since the private method uses Span), if the path is fully formed the method just returns the the original string. You might want to consider using Join instead if the supplied path is already fully formed. Which also accepts Span.

0x90d:
Closing because this refers to outdated VDF Version 2

PR #136, opened 2021-06-05
CommandLine: Status updates write to stderr, can now write JSON to stdout

### LINKED ISSUE #63: Export to json
I'd like to propose an option to export the results to a JSON file. The HTML might be useful to display the results, but it's not very useful when you want to process the results in any other way.

matthewstrasiotto:
Resolves #63


0x90d:
Thanks for your contribution and sorry for not having this reviewed. Closing because this refers to outdated VDF Version 2

Issue #148, opened 2021-08-07
Recognize horizontally flipped videos/Images as duplicate 


Maltragor:
**Environment**
 - OS: Any

**Describe the solution you'd like**
In a certain common domain where a large number of videos are offered (which I won't go into in detail here), it often happens that videos are flipped horizontally.

In order to recognize these as duplicates regardless of whether the scene used as a thumbnail happens to be sufficiently symmetrical, it would be good if VDF compared the grayscale thumbnails with each other in horizontally flipped variants as well. (Probably as a configurable option, since this will increase the scan time).

Maltragor:
up-for-grabs:

I had already thought about it, but then I wasn't sure where to put it in, how much value you put on optimization, and of course I didn't know if you would be interested in integrating/adopting something like that.

How about creating the mirrored variant in ScanForDuplicates in the outer loop for entry and then making a second call with GrayBytesUtils.PercentageDifference(entry_flipped.grayBytes[j], compItem.grayBytes![j]);?

From an optimization point of view, on the other hand, there might be a small advantage in extending PercentageDifference so that three images are passed directly,...
And in principle you could of course also store the mirrored variant in the .db file to not always have to recalculate it. For mirroring itself there are surely more or less performant approaches.
Since you have implemented three variants in PercentageDifference on the one hand, but on the other hand generate e.g. the colored thumbnails by creating files as an intermediate step and also calling ffmpeg multiple times for the individual timestamps per file is certainly not the most performant approach, I could not estimate how important the degree of optimization is to you at which point.

Speaking of which: Why are you actually creating the colored thumbnails as a file? Purely for converting to an Avalonia image, a MemoryStream would do, right? - JoinImages:
```c#
using (MemoryStream memory = new MemoryStream())
{
      img.Save(memory, img.RawFormat);
      memory.Position = 0;
      return new Avalonia.Media.Imaging.Bitmap(memory);
}
```

Since the width of the image is necessary for mirroring, but I think it only occurs somewhere as a ffmpeg parameter (16x16) - should it be introduced somewhere globally as a constant?

0x90d:
Performance is always good. If you believe if you can make something faster without any drawbacks go ahead and make a PR.

Maltragor:
> If you believe if you can make something faster without any drawbacks

That's why I mentioned the coloured thumbnails over files.
[...]

PR #151, opened 2021-08-15
ImageUtils.JoinImages: Added non persistent mode using a memory stream.

Maltragor:
-> Using a memory stream is faster and if the image files is not needed for other purpose there is no reason to store them.

In "Recognize horizontally flipped videos/Images as duplicate #148" I suggested not to create files for displaying the thumbnails, but to convert them into an Avalonia image via memory stream.

You meant:
> If you believe if you can make something faster without any drawbacks go ahead and make a PR.

So here is the small modification. If needed, the files can still be created by calling the function with persistent = true.


Maltragor:
Is it my fault that `AppVeyor build failed`?
Or is there something wrong with your `.github/workflows/dotnetcore.yml` or what ever starts the build/test?

Maltragor:
OK, renaming a branch will close the pull request. learned something new again^^.
Sorry.

I will create  a new PR an copy the comments.

Issue #137, opened 2021-06-09
MKV file: Image resolution used instead of video resolution


StephKoenig:
I've noticed in mkv files which can contain both video and images (for example videos with a cover image) VDF returns the image resolution instead to video resolution. Of course this is not necessarily wrong since VDF can also scan for images but I would suggest if both video and images are found, that the video resolution is used UI presentation.

I would also be handy to have sort by first name (meaning the name of the first file in group). I appreciate for some this might not be useful but for example where you have a folder full of clips where VFD cannot be the sole arbiter say with a 50-80% match setting going through them in order is faster that arbitrarily jumping around the folder.

A 'close' match is much more useful than the standard binary matches especailly when dealing with multipel resolutions and different pre and post-rolls...

Lastly would be handy to have the 'system' menu available on a context click.

~Steph

matthewstrasiotto:
This is really a few separate issues, though I agree with everything you've raised. 

> I've noticed in mkv files which can contain both video and images (for example videos with a cover image) VDF returns the image resolution instead to video resolution. Of course this is not necessarily wrong 

Sounds problematic - I'd argue that this is probably almost always wrong, since no-one really uses the MKV container to store images, and there are plenty of other duplicate finders, that are more mature than this one, that support image comparisons. 

> I would also be handy to have sort by first name 

Sorting by file name would be useful, I agree - in general there needs to be more ways to sort

> A 'close' match is much more useful than the standard binary matches especailly when dealing with multipel resolutions and different pre and post-rolls...

Not sure what this one pertains to 

> Lastly would be handy to have the 'system' menu available on a context click.

What do you mean by that? As in, the right click menu that appears in your file explorer?

StephKoenig:
Thanks for the reply.. great tool

On the MKV mixed mode I agree VDF should only focus on the video tracks and then probably only on the first one (or default one) ... images are very rare and typically only on mkvs if they're used as covers/box art... MetaX and Icaros both provide for cover support creation and display which works really well if interested.

Re: Close match, I can't remember now, maybe I was just saying for my usecase that its more useful than exact - for example a folder full of edits with include/exclude pre and post roll and where you have multiple resolutions... exact match almost never happens and anyway there's plenty of binary comparison tools that can do that already.

Re: System menu.. thats right I meant importing or making available the right click/context click content you might get with your OS on that file type... if its tricky to add withing to a sub menu then maybe a alt-right-clicking can display it solo...

[...]

PR #152, opened 2021-08-15
Bugfix/wrong metadata

Maltragor:
Bugfix for "MKV file: Image resolution used instead of video resolution #137"

If multiple Streams of the same type exists (so e.g. two video streams) VDF displays the media infos of the last stream.
This is probably always or at least often the wrong stream.

In the first commit of this branch I have made only a small change, so VFD uses the first stream of every type for media infos.
(Your `//Workaround if video stream bitrate is not set but in format` also applies the format bitrate to stream 0 only (->`info.Streams[0].BitRate = formatBitrate;`) so my change would fit better to that as well.)

In the second commit I use the default flag of the stream to decide which stream should be used for showing the media infos. 
Since the FFProbeJsonReader was not intended to be easily extended to read a deeper object level, but I need the contents of the inner disposition object, I have rebuilt the reader. Actually there are two variants to choose from. One with JsonSerializer.Deserialize and one based on your variant, which converts the JSON object into a dictionary first.

### DIFF
--- a/VDF.Core/FFTools/FFProbeJsonReader.cs
+++ b/VDF.Core/FFTools/FFProbeJsonReader.cs
@@ -18,21 +18,197 @@ using System;
 using System.Collections.Generic;
 using System.Text.Json;
 using VDF.Core.Utils;
+using System.Text.Json.Serialization;
+
+using DSO = System.Collections.Generic.Dictionary<string, object>; 
+using LO  = System.Collections.Generic.List <object>; 
 
 namespace VDF.Core.FFTools {
+
+#if true
+// ---------------------------------------------------------------------------------
+// V1) JSON -> JsonSerializer.Deserialize -> JsonMediaInfo class -> MediaInfo class
+// ---------------------------------------------------------------------------------
+
 	static class FFProbeJsonReader {
 
-		enum JsonObjects {
-			None,
-			Streams,
-			Format
+		public class JsonMediaInfo
+		{
+			#pragma warning disable CS0649 // Field is never assigned to, and will always have its default
+			public Format? format;
+			public List<Stream>? streams;
+
+			public class Format {
+				public string? bit_rate;
+				public string? duration;
+			}
+			public class Stream	{
+				public string? bit_rate;
+				public int width;
+				public int height;
+				public string? codec_name;
+				public string? codec_long_name;
+				public string? codec_type;
+				public string? channel_layout;
+				public string? pix_fmt;
+				public string? sample_rate;
+				public int index;
+				public string? r_frame_rate;
+				public Disposition? disposition;
+			}
+			public class Disposition {
+				[JsonPropertyName("default")]
+				public int default_;
+			}
+			#pragma warning restore CS0649 // Field is never assigned to, and will always have its default
 		}
 
-		// C# no-alloc optimization that directly wraps the data section of the dll (similar to string constants)
-		// https://github.com/dotnet/roslyn/pull/24621
-		static ReadOnlySpan<byte> StreamsKeyword => new byte[] { 0x73, 0x74, 0x72, 0x65, 0x61, 0x6D, 0x73 }; // = streams
-		static ReadOnlySpan<byte> FormatKeyword => new byte[] { 0x66, 0x6F, 0x72, 0x6D, 0x61, 0x74 }; // = format
-		static ReadOnlySpan<byte> IndexKeyword => new byte[] { 0x69, 0x6E, 0x64, 0x65, 0x78 }; // = index
+		static JsonSerializerOptions serializerOptions = new JsonSerializerOptions
+		{
+			IncludeFields = true,
+		};
+
+		/// <summary>
+		/// Parses FFprobe JSON output and returns a new <see cref="MediaInfo"/>
+		/// </summary>
+		/// <param name="data">JSON output</param>
+		/// <param name="file">The file the JSON output format is about</param>
+		/// <returns><see cref="MediaInfo"/> containing information from FFprobe output</returns>
+		public static MediaInfo Read(byte[] data, string file) {;
+
+			var json = new Utf8JsonReader(data);
+            var jsonInfo = JsonSerializer.Deserialize<JsonMediaInfo>(ref json, serializerOptions);
+
+			var info = new MediaInfo {
+				Streams = new MediaInfo.StreamInfo[jsonInfo?.streams?.Count ?? 0]
+			};
+
+			if (jsonInfo != null && jsonInfo.streams != null) {
+				if (jsonInfo.format?.duration != null && TimeSpan.TryParse(jsonInfo.format.duration, out var duration))
+					/*
+					* Trim miliseconds here as we would have done it later anyway.
+					* Reasons are:
+					* - More user friendly
+					* - Allows an improved check against equality
+					* Cons are:
+					* - Not 100% accurate if you consider a difference of e.g. 2 miliseconds makes a duplicate no longer a duplicate
+					* - Breaking change at the moment of implementation as it doesn't apply to already scanned files
+					*/
+					info.Duration = duration.TrimMiliseconds();
+
+				long.TryParse(jsonInfo.format?.bit_rate, out var formatBitrate);
+		
+				for (int i = 0; i < info.Streams.Count(); i++) {
+					var stream = jsonInfo.streams[i];
+					info.Streams[i] = new MediaInfo.StreamInfo();
+					if (stream.bit_rate != null && long.TryParse(stream.bit_rate , out var bitrate))
+						info.Streams[i].BitRate = bitrate;
+					else if (stream.codec_type == "video")
+						info.Streams[i].BitRate = formatBitrate;
+						//Workaround if video stream bitrate is not set but in format
+
+					info.Streams[i].Width 			= stream.width;
+					info.Streams[i].Height 			= stream.height;
+					info.Streams[i].CodecName 		= stream.codec_name ?? "";
+					info.Streams[i].CodecLongName 	= stream.codec_long_name ?? "";
+					info.Streams[i].CodecType 		= stream.codec_type ?? "";
+					info.Streams[i].ChannelLayout 	= stream.channel_layout ?? "";
+					info.Streams[i].PixelFormat 	= stream.pix_fmt ?? "";
+					info.Streams[i].Default			= stream.disposition?.default_ ?? 0;
+					info.Streams[i].Index 			= stream.index.ToString();
+					
+					if (int.TryParse(stream.sample_rate, out var SampleRate))
+						info.Streams[i].SampleRate = SampleRate;
+
+					if (stream.r_frame_rate != null) {
+						if (stream.r_frame_rate.Contains('/')) {
+							var split = stream.r_frame_rate.Split('/');
+							if (split.Length == 2 && int.TryParse(split[0], out var firstRate) && int.TryParse(split[1], out var secondRate))
+								info.Streams[i].FrameRate = (firstRate > 0 && secondRate > 0) ? firstRate / (float)secondRate : -1f;
+						}
+					}
+				}
+			}
+
+			return info;
+		}
+	}
+
+#else
+
+// ---------------------------------------------------------------------------------
+// V2) JSON -> while(){ Read(json); ...} -> Dictionary/List -> MediaInfo class
+// ---------------------------------------------------------------------------------
+
+	static class FFProbeJsonReader {
+
+		static object? ReadJson(ref Utf8JsonReader json, UInt32 max_depth = 5)
+		{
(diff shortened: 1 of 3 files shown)


VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
 using System.Collections.Generic;
 using System.Text.Json;
 using VDF.Core.Utils;
+using System.Text.Json.Serialization;
+
+using DSO = System.Collections.Generic.Dictionary<string, object>; 
+using LO  = System.Collections.Generic.List <object>;
```
0x90d, line 24: Don't use abbreviations, remove them please

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
 using System.Text.Json;
 using VDF.Core.Utils;
+using System.Text.Json.Serialization;
+
+using DSO = System.Collections.Generic.Dictionary<string, object>; 
+using LO  = System.Collections.Generic.List <object>; 
 
 namespace VDF.Core.FFTools {
+
+#if true
```
0x90d, line 28: Remove this as well

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
 
-		// C# no-alloc optimization that directly wraps the data section of the dll (similar to string constants)
-		// https://github.com/dotnet/roslyn/pull/24621
-		static ReadOnlySpan<byte> StreamsKeyword => new byte[] { 0x73, 0x74, 0x72, 0x65, 0x61, 0x6D, 0x73 }; // = streams
-		static ReadOnlySpan<byte> FormatKeyword => new byte[] { 0x66, 0x6F, 0x72, 0x6D, 0x61, 0x74 }; // = format
-		static ReadOnlySpan<byte> IndexKeyword => new byte[] { 0x69, 0x6E, 0x64, 0x65, 0x78 }; // = index
+		static JsonSerializerOptions serializerOptions = new JsonSerializerOptions
+		{
+			IncludeFields = true,
+		};
```
0x90d, line 69: Can this be readonly?

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
 		/// </summary>
 		/// <param name="data">JSON output</param>
 		/// <param name="file">The file the JSON output format is about</param>
 		/// <returns><see cref="MediaInfo"/> containing information from FFprobe output</returns>
-		public static MediaInfo Read(byte[] data, string file) {
+		public static MediaInfo Read(byte[] data, string file) {;
 
-			var json = new Utf8JsonReader(data, isFinalBlock: false, state: default);
+			var json = new Utf8JsonReader(data);
+            var jsonInfo = JsonSerializer.Deserialize<JsonMediaInfo>(ref json, serializerOptions);
```
0x90d, line 80: Remove the extra spaces

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
+					* - Not 100% accurate if you consider a difference of e.g. 2 miliseconds makes a duplicate no longer a duplicate
+					* - Breaking change at the moment of implementation as it doesn't apply to already scanned files
+					*/
+					info.Duration = duration.TrimMiliseconds();
 
-			var currentStream = -1;
+				long.TryParse(jsonInfo.format?.bit_rate, out var formatBitrate);
+		
+				for (int i = 0; i < info.Streams.Count(); i++) {
+					var stream = jsonInfo.streams[i];
```
0x90d, line 102: Use explicit type

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
+					info.Streams[i] = new MediaInfo.StreamInfo();
+					if (stream.bit_rate != null && long.TryParse(stream.bit_rate , out var bitrate))
```
0x90d, line 104: Use explicit type

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
+					info.Streams[i].Height 			= stream.height;
+					info.Streams[i].CodecName 		= stream.codec_name ?? "";
+					info.Streams[i].CodecLongName 	= stream.codec_long_name ?? "";
+					info.Streams[i].CodecType 		= stream.codec_type ?? "";
+					info.Streams[i].ChannelLayout 	= stream.channel_layout ?? "";
+					info.Streams[i].PixelFormat 	= stream.pix_fmt ?? "";
+					info.Streams[i].Default			= stream.disposition?.default_ ?? 0;
+					info.Streams[i].Index 			= stream.index.ToString();
+					
+					if (int.TryParse(stream.sample_rate, out var SampleRate))
```
0x90d, line 120: Use explicit type

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
-						currentObject = JsonObjects.Streams;
-						break;
-					}
-					if (valueSpan.SequenceEqual(FormatKeyword)) {
-						currentObject = JsonObjects.Format;
-						break;
+					if (stream.r_frame_rate != null) {
+						if (stream.r_frame_rate.Contains('/')) {
+							var split = stream.r_frame_rate.Split('/');
+							if (split.Length == 2 && int.TryParse(split[0], out var firstRate) && int.TryParse(split[1], out var secondRate))
```
0x90d, line 126: Use explicit types. Why do you check for split length to be 2 when you are using only the first one?

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
+			var json = new Utf8JsonReader(data);
+            var jsonInfo = JsonSerializer.Deserialize<JsonMediaInfo>(ref json, serializerOptions);
 
+			var info = new MediaInfo {
+				Streams = new MediaInfo.StreamInfo[jsonInfo?.streams?.Count ?? 0]
+			};
 
-			var streams = new List<Dictionary<string, object>>();
-			var format = new Dictionary<string, object>();
+			if (jsonInfo != null && jsonInfo.streams != null) {
```
0x90d, line 86: This can be merged

VDF.Core/FFTools/FFProbeJsonReader.cs
```diff
@@ -18,114 +18,212 @@
...
+		/// <param name="file">The file the JSON output format is about</param>
+		/// <returns><see cref="MediaInfo"/> containing information from FFprobe output</returns>
+		public static MediaInfo Read(byte[] data, string file) {
+			var json = new Utf8JsonReader(data, isFinalBlock: false, state: default);
+		
+			//json.Read();
+			DSO root = (DSO)(ReadJson(ref json) ?? new DSO());
+			
+			LO  streams = root.ContainsKey("streams") ? (LO)(root["streams"]) : new LO();
+			DSO format  = root.ContainsKey("format")  ? (DSO)(root["format"]) : new DSO();
```
0x90d, line 226: Don't use abbreviations

VDF.Core/MediaInfo.cs
```diff
@@ -64,6 +64,8 @@ public class StreamInfo {
 			[ProtoMember(11)]
 			public float FrameRate { get; set; }
 
+			[ProtoMember(12)]
+			public int Default { get; set; }
```
0x90d, line 68: This is a very bad name for a variable. Also can you explain why we need this?

0x90d (changes requested):

VDF.Core/FFTools/FFProbeJsonReader.cs
Maltragor: I was so happy when I found a way to shorten this because `typedef` doesn't seem to exist in C#.
Do you generally not define your own types for such purposes in C#, or is only the variant with using not the best way?

VDF.Core/FFTools/FFProbeJsonReader.cs
Maltragor: *smirk*
This is your code, see your version, line 177.
>  check for split length to be 2 
Because in the last part of the line,` int.TryParse(split[1], out var secondRate)` is used, right?

VDF.Core/FFTools/FFProbeJsonReader.cs
Maltragor: There are two versions of FFProbeJsonReader implementation.
Removing the ` #if ... #else ... #endif` would not work.

=> I offered you the two variants so you can choose which one you like better. I didn't know if there was a reason that you didn't use the ready deserialization function in your version.
And if there are still problems with one of the variants, you would still have the other to choose from.
Therefore, one would either delete one of the two variants completely, or save it just in case.

VDF.Core/FFTools/FFProbeJsonReader.cs
Maltragor: Since I was programming in C# for the first time as part of these commits, optimizing a null query was not a top priority.^^
Would (jsonInfo?.streams != null) be the appropriate thing to do, or how would you merge it?

VDF.Core/MediaInfo.cs
Maltragor: There seems to be something reserved `default` I noticed when I created the serialization class (but not jet read what's behind the default)

the problem to fix is to read the metadata of the one that ffmpeg displays / uses to create the thumbnails, if there is more than one video stream.

In the metadata of the stream there is the `disposition` block, in which there is a field `default` (furthermore also fields like `attached_pic`,..).
In the video examples I had seen, the relevant stream / the stream to be played was also flagged with default=1.
So my assumption was that ffmpeg also selects this stream if you don't specify one specifically.
=> My change reads the `default` flag, stores it in the `MediaInfo` (under the name `Default`) and uses the information to select the stream.

Maltragor: But just now I came across the following page:
https://ffmpeg.org/ffmpeg.html#Automatic-stream-selection

Of course, I don't know if that's still current. Just as I don't know in which video container formats the disposition section exists... or however ffprobe creates it.

But back to the description I found: 

> 4.1.1 Automatic stream selection
> 
> In the absence of any map options for a particular output file, ffmpeg inspects the output format to check which type of streams can be included in it, viz. video, audio and/or subtitles. For each acceptable stream type, ffmpeg will pick one stream, when available, from among all the inputs.
> 
> It will select that stream based upon the following criteria:
> 
>     for video, it is the stream with the highest resolution,
>     for audio, it is the stream with the most channels,
>     for subtitles, it is the first subtitle stream found but there’s a caveat. The output format’s default subtitle encoder can be either text-based or image-based, and only a subtitle stream of the same type will be chosen. 
> 
> In the case where several streams of the same type rate equally, the stream with the lowest index is chosen. 

**Based on this, my extension to read the default flag would not be necessary because it is not the feature used by ffmpeg.** (second commit on this branch)

Instead, you would extend the first commit in this branch (use stream with lowest index) with the resolution criterion mentioned in the description and be done with it.
_(I will do this within the next days, its only a small local change... extended if condition + last resolution variable)_

This would then eliminate the need to use my modified JsonReader. Except of course, you like one of my two versions better.... But it would be also not bad, if not.

VDF.Core/FFTools/FFProbeJsonReader.cs
0x90d: I don't use abbreviations. If you see them in code, especially in code you haven't written yourself. You have to look up what they mean. And the more abbreviations and the more complex the code, the harder it can be read.

VDF.Core/FFTools/FFProbeJsonReader.cs
0x90d: Sorry I missed this one, you're right.

VDF.Core/FFTools/FFProbeJsonReader.cs
0x90d: Yes

VDF.Core/MediaInfo.cs
0x90d: This is very interesting, nice find!

Maltragor:
See _Selection rules for displayed MediaInfo adapted to the default rules_ … #169 

=> The extension to read deeper levels of the json object is no longer necessary, so this PR is basically done here and can be rejected.

If you are still interested in the alternative implementations of the reader, I can still make the adjustments you noted. (Otherwise just close the PR)

PR #153, opened 2021-08-15
Feature/h flip

Maltragor:
Feature:  "Recognize horizontally flipped videos/Images as duplicate #148 "

Based on my implementation proposal in #148 I extended ScanForDuplicates() to compare the horizontally flipped grayscale images as well. 

Also, since this means the image width is now needed in multiple files, I extended ScanEngine and GrayScaleUtils so that you pass the width to the respective functions as parameters.
In principle, one could also implement a changeable thumbnail size on the basis, or use other resolutions for the grayscale images. But this was not the goal of the change and the optimized GrayScale functions would support this only limitedly.
The search for flipped media is configurable in principle and the information whether it is a flipped duplicate is also available. However, the connection to the GUI is completely missing. Here I had the hope that you would take over this, since I have no idea about Avalonia at all and would also have to learn extra there. 

The first commit includes all necessary changes. The second one improves the performance of the flipping.
Because I had to test an debug the Avx2 and Sse2 things, I wrote a test function and different flip functions combined by a wrapper function. (this is also the reason why I split your PercentageDifference)
Of course the test function can be removed if you get no benefit from it.


VDF.Core/ScanEngine.cs
```diff
@@ -214,35 +214,53 @@ public sealed class ScanEngine {
...
+						do {
+							if (entry.IsImage) {
+								percent[0] = GrayBytesUtils.PercentageDifference(grayBytes[flip][0]!, compItem.grayBytes[0]!);
+								isDuplicate = (percent[0] < percentageDifference);
+							}
+							else {
+								for (var j = 0; j < positionList.Count; j++) {
+									percent[j] =
+										GrayBytesUtils.PercentageDifference(grayBytes[flip][entry.GetGrayBytesIndex(positionList[j])]!, compItem.grayBytes[compItem.GetGrayBytesIndex(positionList[j])]!);
+									if (!(percent[j] < percentageDifference)) {
```
0x90d, line 249: Just invert, the way its written its hard to read

VDF.Core/ScanEngine.cs
```diff
@@ -214,35 +214,53 @@ public sealed class ScanEngine {
...
+										isDuplicate = false;
+										break; 
+									}
 								}
-								else { break; }
 							}
-						}
-
+						} while(!isDuplicate && Settings.CompareHorizontallyFlipped && flip++ == 0);
```
0x90d, line 255: This is also hard to read. It takes a while to understand what the loop does.

VDF.Core/ScanEngine.cs
```diff
@@ -296,9 +314,9 @@ public sealed class ScanEngine {
 							try {
 								Image bitmapImage = Image.FromFile(entry.Path);
 								float resizeFactor = 1f;
-								if (bitmapImage.Width > 100 || bitmapImage.Height > 100) {
-									float widthFactor = bitmapImage.Width / 100f;
-									float heightFactor = bitmapImage.Height / 100f;
+								if (bitmapImage.Width > thumbnailWidth || bitmapImage.Height > thumbnailWidth) {
+									float widthFactor = bitmapImage.Width / (float)thumbnailWidth;
+									float heightFactor = bitmapImage.Height / (float)thumbnailWidth;
```
0x90d, line 319: I cannot see why this change is necessary

VDF.Core/ScanEngine.cs
```diff
@@ -342,7 +361,7 @@ public sealed class ScanEngine {
 			ThumbnailsRetrieved?.Invoke(this, new EventArgs());
 		}
 
-		static void GetGrayBytesFromImage(FileEntry imageFile) {
+		static void GetGrayBytesFromImage(FileEntry imageFile, int width) {
```
0x90d, line 364: I cannot see why this change is necessary

VDF.Core/Settings.cs
```diff
@@ -36,5 +36,7 @@ public sealed class Settings {
 		public float Percent = 96f;
 
 		public int ThumbnailCount = 1;
+
+		public bool CompareHorizontallyFlipped = true;
```
0x90d, line 40: It should be false by default

VDF.Core/Utils/GrayBytesUtils.cs
```diff
@@ -38,7 +37,7 @@ static class GrayBytesUtils {
 			return 100d / data.Length * darkPixels < darkProcent;
 		}
 
-		public static unsafe byte[]? GetGrayScaleValues(Bitmap original, double darkProcent = 80) {
+		public static unsafe byte[]? GetGrayScaleValues(Bitmap original, int width, double darkProcent = 80) {
```
0x90d, line 40: I cannot see why this change is necessary

VDF.Core/Utils/GrayBytesUtils.cs
```diff
@@ -68,38 +67,235 @@ static class GrayBytesUtils {
...
+			Span<Vector128<byte>> vImg_flipped = MemoryMarshal.Cast<byte, Vector128<byte>>(flip_img);
+			Span<Vector128<byte>> vFlipp_shuf = MemoryMarshal.Cast<byte, Vector128<byte>>(flipp_shuf256);
+
+			for (int i = 0; i < vImg.Length; i++)
+				vImg_flipped[i] = Avx2.Shuffle(vImg[i], vFlipp_shuf[0]);
+
+			return flip_img;
+		}
+
+#if false
```
0x90d, line 186: Remove this

VDF.Core/Utils/GrayBytesUtils.cs
```diff
@@ -68,38 +67,235 @@ static class GrayBytesUtils {
...
+	//-------------------------------------------------------------------------
+	// Unit / performance tests
```
0x90d, line 188: Performance tests should be done with benchmarkdotnet and should be moved into a new project

VDF.Core/ViewModels/DuplicateItem.cs
```diff
@@ -24,10 +24,18 @@ public class DuplicateItem  {
...
 		public DuplicateItem() { }
 #pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
-		public DuplicateItem(FileEntry file, float percent, Guid groupID) {
+
+		[Flags]
+		public enum DuplicateFlags : short
+		{
+			None = 0,
+			Flipped = 1,
+		};
```
0x90d, line 33: Move into new file

VDF.Core/FFTools/FfmpegEngine.cs
```diff
@@ -29,7 +29,9 @@ static class FfmpegEngine {
 		public static byte[]? GetThumbnail(FfmpegSettings settings) {
 			using var process = new Process {
 				StartInfo = new ProcessStartInfo {
-					Arguments = $" -hide_banner -loglevel panic -y {(UseCuda ? "-hwaccel cuda" : string.Empty)} -ss {settings.Position} -i \"{settings.File}\" -t 1 -f {(settings.GrayScale == 1 ? "rawvideo -pix_fmt gray" : "mjpeg")} -vframes 1 {(settings.GrayScale == 1 ? "-s 16x16" : "-vf scale=100:-1")} \"-\"",
+					Arguments = $" -hide_banner -loglevel panic -y {(UseCuda ? "-hwaccel cuda" : string.Empty)} -ss {settings.Position} -i \"{settings.File}\" -t 1 " +
+					            $"-f {(settings.GrayScale == 1 ? "rawvideo -pix_fmt gray" : "mjpeg")} -vframes 1 " +
+								$"{(settings.GrayScale == 1 ? $"-s {settings.Width}x{settings.Width}" : $"-vf scale={settings.Width}:-1")} \"-\"",
```
0x90d, line 34: Either use @ for multi line or make it single line again.

0x90d (changes requested):

VDF.Core/ScanEngine.cs
```diff
@@ -214,35 +214,53 @@ public sealed class ScanEngine {
 					while (pauseTokenSource.IsPaused) Thread.Sleep(50);
 
 					var entry = ScanList[i];
+					float[] percent = new float[positionList.Count];
+					Dictionary<double, byte[]?>[] grayBytes = new Dictionary<double, byte[]?>[2];
```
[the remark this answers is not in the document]
Maltragor: The base idea of crating the array was to combine the two cases (flip or not) retrospectively by a simple loop so that the actual functionality remains clear and the two cases do not have to be distinguished again.

And yes, it would be possible to replace the `2` by `(Settings.CompareHorizontallyFlipped ? 2 : 1)`
But this has no effect on the image case. In the image case, in principle you would not need a dictionary. But then image an video case would need to be completely separated.

VDF.Core/ScanEngine.cs
Maltragor: Of course I can invert it if you like, but it was intentional.
Because some lines above: `isDuplicate = (percent[0] < percentageDifference);`, so it says: set to true, if smaller.
So I prefered to keep the condition with the smaller but negate it, because isDuplicate is set to false.

_But that is probably a matter of taste. I only wrote it, because initially I inverted the condition, so used ">=" but then thought it's easier to understand the other way around._

VDF.Core/ScanEngine.cs
Maltragor: That was my way of merging a condition :-)

But probably you would think about something like: (?)
```c#
for (flip = 0; flip < Settings.CompareHorizontallyFlipped ? 2 : 1; flip++)
{
 ...
if (isDuplicate)
     break;
}
```

VDF.Core/ScanEngine.cs
Maltragor: For flipping itself it is not necessary, that's right.

But:
- The ffmpeg command specified the 16x16 size as string.
- I didn't like it when now in more places in other files a hardcoded assumption is made to the 16x16 size. Especially since e.g. also a feature request exists regarding parameterizable ffmpeg parameters.
- So I thought that the 16x16 size should be specified centrally in one place. But since both the 100:-1 scaling of the thumbnails and the 16:16 of the grayscale images run via one function, both were related to each other
- And on the other hand there is the tooltip zoom function of the thumbnails with a bad quality based on the low thumbnail resolution.
=> So I come to the idea, it would be the best solution, if the thumbnail and graybyte size would be parametrizable.
On the one hand it solves my original problem, having a clear central point of defining the size while on the other hand it would be the base of a future user configurable thumbnail size or a better zoom (not scaling tooltip image up but scaling thumbnail down based on tooltip image).
That is why I consistently exchanged  the thumbnail / graybyte size constants with the parameters.

Maltragor: I don't comment the other places which also refer to this because it would be the same answer.
-> Central place of definition + flexibility to easily support other resolutions in future.
Maybe also for other graybyte resolutions, but this was not on focus. (So no focus on optimisation other resolutions)

VDF.Core/Utils/GrayBytesUtils.cs
Maltragor: Do you have a place for unit / performance / ... tests?
Yes, the file is not the best place, but I didn't want to keep the code section from you, in case you also wanted to try out the variants in comparison.

VDF.Core/Utils/GrayBytesUtils.cs
Maltragor: And setting that up would be too much of a good thing, at least from my side at the moment.
I had Visual Studio *Code* set up to debug a .NET project.... but that was enough for my goal.

But of course if there would be an existing place for such things I would use it.

VDF.Core/FFTools/FfmpegEngine.cs
Maltragor: OK.
But what is bad about splitting a very long string by using `+` in order to make it readable?

Maltragor:
You added a lot of comments to small or non functional things but you have not yet said anything about what I consider to be the essential issues.
So is the way I implemented the feature OK for you.
Would you do the GUI connection. (E.g. configurability of the setting (*1), adding a "flipped" to the % view,...)
... or we are still in a phase in which this is not clear at all and after the small stuff is then possibly adjusted in several iterations, the feedback could still come that it does not fundamentally fit^^
Or applies that everything to which nothing was noted OK.

(*1) Only as short comment. I will open an issue at some point.
Saving and managing settings seems to be tightly coupled to the GUI, which I don't find so ideal.
It would make more sense from my point of view, if this is rather placed in the core part and the application/GUI only calls a load/store and reads/writes the members of the settings class.
It would then be easier to add settings that only exist in the file, or to move the graphical part to a separate dialog if the number of settings increases even more....

VDF.Core/ScanEngine.cs
0x90d: Yeah

VDF.Core/Utils/GrayBytesUtils.cs
0x90d: I dont have a place yet. Its still on my very long to do list.

VDF.Core/FFTools/FfmpegEngine.cs
0x90d: I'm not sure if the compiler merges it into one string or if its getting joined at runtime, if you use `+`, which costs performance.

VDF.Core/ScanEngine.cs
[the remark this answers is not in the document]
Maltragor: I have now taken a different approach.
Instead of the loop and the additional array, the comparison is now outsourced to a function. 
This is the clearest way from my point of view.

VDF.Core/Utils/GrayBytesUtils.cs
Maltragor: Removed

VDF.Core/ViewModels/DuplicateItem.cs
Maltragor: Done -> DuplicateFlags.cs

VDF.Core/FFTools/FfmpegEngine.cs
Maltragor: again one line

Maltragor:
Modifications:
- Splitting two-step search as function instead of loop + array
- Checkbox for enabling flipped search + XML load/save added to UI
- Flipped flag is displayed in UI (*1)

(*1) I have added it to the similarity column. But this has the effect that clicking on the Header no longer sorts the column. I don't know if it is possible to keep the column sorting, so if it was a imported feature, I could only remove the flag again in order to get sorting back.

I have stumbled in a few places over what I see as confusing / inconsistent naming in terms of percent and normalized value and difference vs similarity.
e.g. _percentageDifference = 1.0f - Settings.Percent / 100f_
or  `percSame = percent.Average();` (bit is is the difference between [0..1])  -> percent - > Similarity = (1f - percent) * 100;
When I found such places while modifying a function, I renamed the variables. I hope it's OK, otherwise I undo it again.

VDF.Core/FFTools/FfmpegEngine.cs
```diff
@@ -68,7 +68,8 @@ static class FfmpegEngine {
 				var data = GetThumbnail(new FfmpegSettings {
 					File = videoFile.Path,
 					Position = TimeSpan.FromSeconds(position),
-					GrayScale = 1
+					GrayScale = 1,
+					Width = width,
```
0x90d, line 72: I think that Width should be hardcoded again and not a setting/parameter. I'm not commenting other places where you changed it.

VDF.Core/ScanEngine.cs
```diff
@@ -199,21 +201,63 @@ public sealed class ScanEngine {
...
+					double idx = entry.GetGrayBytesIndex(positionList[j]);
+					flippedGrayBytes.Add(idx, GrayBytesUtils.FlipGrayScale(entry.grayBytes[idx]!, grayScaleWidth));
+				}
 
-			var percentageDifference = 1.0f - Settings.Percent / 100f;
-			var duplicateDict = new Dictionary<string, DuplicateItem>();
+			return flippedGrayBytes;
+		}
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
```
0x90d, line 229: I don't think this will be inlined because the method body is too big. AggressiveInlining is just a suggestion for the compiler, you cannot enforce it.

VDF.Core/ScanEngine.cs
```diff
@@ -228,37 +272,27 @@ public sealed class ScanEngine {
...
-						float[] percent;
-						if (entry.IsImage) {
-							percent = new float[1];
-							percent[0] = GrayBytesUtils.PercentageDifference(entry.grayBytes[0]!, compItem.grayBytes[0]!);
-							if (percent[0] < percentageDifference)
-								duplicateCounter++;
+						FileEntry compItem = ScanList[n];
+						DuplicateFlags flags = DuplicateFlags.None; 
+						bool dupplicate = isDuplicate(entry, null, compItem, differenceLimit, out var difference);
+						if (!dupplicate && Settings.CompareHorizontallyFlipped)
```
0x90d, line 283: First condition is false when one item is an image and the other is a video file. In this case you will call "isDuplicate" again which isn't needed.

VDF.Core/ScanEngine.cs
```diff
@@ -228,37 +272,27 @@ public sealed class ScanEngine {
...
-						var duplicateCounter = 0;
-						float[] percent;
-						if (entry.IsImage) {
-							percent = new float[1];
-							percent[0] = GrayBytesUtils.PercentageDifference(entry.grayBytes[0]!, compItem.grayBytes[0]!);
-							if (percent[0] < percentageDifference)
-								duplicateCounter++;
+						FileEntry compItem = ScanList[n];
+						DuplicateFlags flags = DuplicateFlags.None; 
+						bool dupplicate = isDuplicate(entry, null, compItem, differenceLimit, out var difference);
```
0x90d, line 282: the boolean name should be "isDuplicate"

VDF.Core/ScanEngine.cs
```diff
@@ -199,21 +201,63 @@ public sealed class ScanEngine {
...
+					flippedGrayBytes.Add(idx, GrayBytesUtils.FlipGrayScale(entry.grayBytes[idx]!, grayScaleWidth));
+				}
 
-			var percentageDifference = 1.0f - Settings.Percent / 100f;
-			var duplicateDict = new Dictionary<string, DuplicateItem>();
+			return flippedGrayBytes;
+		}
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		bool isDuplicate(FileEntry entry1, Dictionary<double, byte[]?>? grayBytes1, FileEntry entry2, float differenceLimit, out float difference)
```
0x90d, line 230: Method name should be changed to something like "checkIfDuplicate"

VDF.Core/Settings.cs
0x90d: Again this should be false by default

VDF.Core/Utils/GrayBytesUtils.cs
```diff
@@ -68,38 +67,121 @@ static class GrayBytesUtils {
...
-			long diff = 0;
-			if (Avx2.IsSupported) {
-				Vector256<ushort> vec = Vector256<ushort>.Zero;
-				Span<Vector256<byte>> vImg1 = MemoryMarshal.Cast<byte, Vector256<byte>>(img1);
-				Span<Vector256<byte>> vImg2 = MemoryMarshal.Cast<byte, Vector256<byte>>(img2);
 
-				for (int i = 0; i < vImg1.Length; i++)
-					vec = Avx2.Add(vec, Avx2.SumAbsoluteDifferences(vImg2[i], vImg1[i]));
+			if (Avx2.IsSupported && img1.Length % 32 == 0)
+				return PercentageDifferenceAvx2(img1, img2);
```
0x90d, line 76: Please inline the code again and not make it a call. Also why did you added % 32 == 0? This is always be the case.

VDF.Core/Utils/GrayBytesUtils.cs
```diff
@@ -68,38 +67,121 @@ static class GrayBytesUtils {
...
+
+
+	// Different FlipGrayScale versions:
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		private static byte[] FlipGrayScaleReverse(byte[] img, int img_width)
+		{
+			int rows = img.Length / img_width;
+			byte[] flip_img = (byte[])img.Clone();
+			for (int r = 0; r < rows; r++)
```
0x90d, line 150: rename r to i

VDF.GUI/ViewModels/MainWindowViewModel.cs
```diff
@@ -553,6 +562,7 @@ public class MainWindowViewModel : ReactiveObject {
 			Scanner.Settings.Percent = Percent;
 			Scanner.Settings.MaxDegreeOfParallelism = MaxDegreeOfParallelism;
 			Scanner.Settings.ThumbnailCount = Thumbnails;
+			Scanner.Settings.CompareHorizontallyFlipped =CompareHorizontallyFlipped;
```
0x90d, line 565: typo, missing a space

VDF.GUI/Views/MainWindow.xaml
```diff
@@ -511,10 +512,34 @@
...
+                                                    <TextBlock
+                                                        Grid.Row="2"
+                                                        Margin="2"
+                                                        IsVisible="{Binding ItemInfo.Flags, Converter={StaticResource IsFlaggedConverter}}"
+                                                        Text="{Binding ItemInfo.Flags}"
+                                                        TextWrapping="Wrap" />
+                                                </Grid>
+                                            </DataTemplate>
+                                        </DataGridTemplateColumn.CellTemplate>
+                                    </DataGridTemplateColumn >
```
0x90d, line 542: I don't think this is necessary. The preview image already reveals if its a flipped match or not.

0x90d (changes requested):

VDF.GUI/Views/MainWindow.xaml
Maltragor: OK (I'll interpret it as: Please revert the change.)

- Is it ok to keep `StringFormat=\{0:0.##\}` or did you intentionally display so many decimal places?
- If not showing the flipped flag, ItemInfo.Flags has in principle no more use at this moment. So does your request also includes the removal of the flag at all?

_My thoughts on this were that the specification of the percentage similarity on its own no longer has the full significance. Or that one recognizes with many duplicate findings the flipped property then also more simply,...
The flag was also only displayed if it is != None in order not to disturb in the normal case.
But of course the basic statement is right, that if you look at the pictures yourself, you should recognize the flipping in the rule._

VDF.Core/Utils/GrayBytesUtils.cs
Maltragor: OK
_(Even if I found the division in principle nicer, because it has allowed to test also specifically the individual variants.)_

> added % 32 == 0
Because of my approach to make the width configurable. So as you requested not to do this, this is one of the  _"I'm not commenting other places where you changed it."_ 
As I had mentioned somewhere before, as I made `width` a parameter my focus was not to support the best performance for any other resolutions but not to break the functions if a unfavorable resolution would be selected.
And because of the 32 Byte MemoryMarshal mapping I had the feeling it could be a problem if the underlying Byte array is smaller then this region.

VDF.Core/ScanEngine.cs
Maltragor: Was this only a hint or should it be removed.
(As I added it, I used PercentageDifference as a reference. But is this function small enough to get inlined?)

VDF.Core/FFTools/FfmpegEngine.cs
Maltragor: Too bad,... but OK.
_Then there is probably no user configurable thumbnail size or a better tooltip zoom planed in near future.
Whereby I don't really need both,... the parameterizable approach would have just allowed it. Or should one make the ffmpeg parameters freely configurable once, one could have docked that also there to adjust the image size with.
At least that were just so my thoughts behind it,..._

VDF.Core/ScanEngine.cs
Maltragor: Two questions:
1) https://github.com/0x90d/videoduplicatefinder/blob/364b094355bf2c351ba1de1b3a417ac4f3a329c0/VDF.Core/ScanEngine.cs#L233 I had taken over from you.
But as we just talk about it. Is this really what we want?
If entry is a video it would still be compared with a compItem-Image. 
So shouldn't it be  `if (entry.IsImage != compItem.IsImage) continue; `

_2) Because _isDuplicate()_ contains the complete comparison the image/video check is included there, but right, so the function is called in principle unnecessarily a second time.
So if "unnecessarily" calls should be reduced at all, the image/video check would need to be placed outside the function at the beginning oft the loop. This would also prevent the first "unnecessarily" call. But in consequence the image/video check inside the function then becomes unnecessary - but removing the check from the function would make the function a bit incomplete without the context it is called from....
Probably too many thoughts on this topic, I know....
So if you say nothing else about it I just add the image/video check to line 283 requested. (After (1) is clarified)_

Maltragor:
- Commented undesirable changes reverted
- Renaming,... done

The Condition based on original:
https://github.com/0x90d/videoduplicatefinder/blob/364b094355bf2c351ba1de1b3a417ac4f3a329c0/VDF.Core/ScanEngine.cs#L233
is changed to `if (entry1.IsImage != entry2.IsImage)`. The original condition only did not crash because `GetGrayBytesIndex()`  always returns 0 if duration is 0 what applies to all images. But it leads to a indeterministic mixture of videos an images within a duplicate group depending on the order of the file list / file names. (... if a low difference limit is selected to be detected as similar)

0x90d:
@Maltragor I'm closing this since it was heavily outdated. If you can update it with current code in master I will merge it. You can ignore my previous comments here.

Maltragor:
OK. 
I had after the renewed review request in august first waited a little whether you still get in touch,... and then when I thought to myself,... actually I would now have to ask again whether you still respond to it - it was it from my side temporally unfavorable.... 
And exactly... meanwhile a lot has happened in the master.
I had some time ago for me again locally rebased the branch, but also the state is no longer current.
Therefore, it was still on my to-do list to ask whether you are still interested in the PR.
Nice that you have written something about it after the close :-)

Then I update it the days again and create a new PR.
I had taken most of the comments into account in the last update, or so I thought/hoped^^.

0x90d:
I thought about this for quite a long time. At first I thought it would be a niche feature but the more I think about, the more I like the addition. Many duplicate finders don't have this feature and when it's optional it allows people to find more duplicates if they want to.

I also thought about adding flipped graybytes to database. This is why I have ignored your PR for quite a while. But in the end no one knows if we don't add support for any other rotation in the future and it would just blow up the database creating entries for all of them.

So I will go with your approach, it seems to be the best way as of now. I just cannot merge it because of the massive conflicts so I do appreciate if you could re-create it.

0x90d:
FYI: There is just one thing I've noticed when I played with flipped duplicates. The lower the similarity percentage is the higher the chance VDF can already detected flipped duplicates. In my tests a flipped duplicate was detected with a similarity of 86%. But with scanned for flipped duplicates the similarty raised up to 98%.

My question is we want to skip flipped check if a duplicate is found or if we still check it and take whatever generates higher similarity. The reason is that people may run some rules to drop anything below X % similarity. In my case the flipped version would always end with 86% score in your implementation while it could have 98%.

And second question is IF we want to check for both ways if we make it optional or not. In my opinion, because flipped check already takes much longer, I wouldn't make it optional. I would rather want to have the best results.

Maltragor:
> The reason is that people may run some rules to drop anything below X % similarity.

This is a point I had not considered. Because I don't use the filter or sorting and manually check every listed video. And most of the time I do several passes with different Similarity limits instead. And when reaching the lower similarity levels there are many large groups with rather completely different videos that I only scroll over it quickly at the end to find maybe one or the other actual duplicate. (e.g. videos with a small time offset).
But yes, since there are usually far fewer duplicates than different videos in relation to the entire video collection, it no longer has a major impact on the scan time if you always compare the flipped variant even when the non flipped was similar.
Except if you set the similarity threshold very very low and thus get extremely many false positives. But then the duration already increases significantly just by creating all the thumbnails...
From there I share your assessment here and would also not offer any further configuration setting. You would then also have to explain to the user under which conditions he has any (small) advantage/disadvantage from the setting.

Maltragor:
> The lower the similarity percentage is the higher the chance VDF can already detected flipped duplicates.

I know, in the past (VDF2) there was a time when I thought VFD supports flipped videos. But later i realized it was just a coincidence depending on the scene symmetry and a sufficiently low threshold.
Setting a low threshold could also help to find videos wit time offsets. But of course only if you catch a scene that barely changes in both videos over a sufficient amount of time.

PR #156, opened 2021-08-16
ImageUtils.JoinImages: Added non persistent mode using a memory stream.

Maltragor:
-> Using a memory stream is faster and if the image files is not needed for other purpose there is no reason to store them.

In "Recognize horizontally flipped videos/Images as duplicate #148" I suggested not to create files for displaying the thumbnails, but to convert them into an Avalonia image via memory stream.

You meant:

    If you believe if you can make something faster without any drawbacks go ahead and make a PR.

So here is the small modification. If needed, the files can still be created by calling the function with persistent = true.

### DIFF
--- a/VDF.GUI/Utils/ImageUtils.cs
+++ b/VDF.GUI/Utils/ImageUtils.cs
@@ -22,10 +22,8 @@ using VDF.Core.Utils;
 namespace VDF.GUI.Utils {
 	static class ImageUtils {
 		public static string ThumbnailDirectory => Core.Utils.FileUtils.SafePathCombine(CoreUtils.CurrentFolder, "Thumbnails");
-		public static Bitmap JoinImages(List<System.Drawing.Image> pImgList) {
+		public static Bitmap JoinImages(List<System.Drawing.Image> pImgList, bool persistent = false) {
 			if (pImgList == null || pImgList.Count == 0) return null;
-
-			string file = Core.Utils.FileUtils.SafePathCombine(ThumbnailDirectory, Path.GetRandomFileName() + ".jpeg");
 
 			int height = pImgList[0].Height;
 			int width = 0;
@@ -41,8 +39,20 @@ namespace VDF.GUI.Utils {
 				}
 				g.Save();
 			}
-			img.Save(file, img.RawFormat);
-			return new Bitmap(file);
+			
+			if (persistent) {
+				string file = Core.Utils.FileUtils.SafePathCombine(ThumbnailDirectory, Path.GetRandomFileName() + ".jpeg");
+				img.Save(file, img.RawFormat);
+				return new Bitmap(file);
+			}
+			else {
+				using (MemoryStream memory = new MemoryStream())
+				{
+					img.Save(memory, img.RawFormat);
+					memory.Position = 0;
+					return new Avalonia.Media.Imaging.Bitmap(memory);
+				}
+			}			
 		}
 	}
 }


Maltragor:
Is it my fault that AppVeyor build failed?
Or is there something wrong with your .github/workflows/dotnetcore.yml or what ever starts the build/test?

0x90d:
> Or is there something wrong with your .github/workflows/dotnetcore.yml

Yes, I haven't updated it for v3 yet

VDF.GUI/Utils/ImageUtils.cs
```diff
@@ -41,8 +39,20 @@ static class ImageUtils {
...
-			img.Save(file, img.RawFormat);
-			return new Bitmap(file);
+			
+			if (persistent) {
+				string file = Core.Utils.FileUtils.SafePathCombine(ThumbnailDirectory, Path.GetRandomFileName() + ".jpeg");
+				img.Save(file, img.RawFormat);
+				return new Bitmap(file);
+			}
+			else {
+				using (MemoryStream memory = new MemoryStream())
```
0x90d, line 49: Simply the using

0x90d:
I don't know how exact Avalonia works. My assumption was that if it stored as file. Then Avalonia loads the image from disk only when its visible on the screen. When created in memory its always there. This can be a huge difference for people with small or large memory, at least if my assumption is correct.

Can we make this a setting?

VDF.GUI/Utils/ImageUtils.cs
Maltragor: I use this position as a representative for others of the type.

Let me put it this way:
From my side it would be completely OK if you adjust such little things yourself. So in the sense that it's not a problem for me if you only take parts, see my code only as a suggestion or especially non-functional adjustments without major coordination simply do it yourself.
At least in terms of time, it costs us both more if you write something that you already have a concrete idea of how it should be, I then sometimes have to find out what you might want to change, I then make the change that you want.... and three weeks later the line is finally so that you accept it.

Maltragor:
But JoinImages creates the file each time it is called. So when the expectation is that JoinImages is called each time a table item becomes visible, this would also be not the best way. And on the other hand, JoinImages needs the thumbnails as parameter so they must exist in RAM or have to be extracted from the video each time,.... 
So in my opinion such a memory saving feature should be set up as follows:
- Extract Thumbnails from video
- Join and create file (no bitmap as return value but e.g. a UUID)
- If a duplicate row item becomes visible the file is loaded
Maybe i have missed the code location where the file is loaded again,... but as the filename is `Path.GetRandomFileName()` and not returned, I don't thinks so.

RAM usage: (only a train of thought / an estimation)
Assume the image is 100x100, unpacked 3 byte/pixel =  30k; 5 thumbnails are configured = 150k; 10.000 duplicates found = 1.5 GB (round about... 1000 vs 1024 doesn't matter).
10.000 duplicates (not video collection size) should be more then worst case... probably you should split your search... it would take about three hours if taking one second for looking at each duplicate,...
And 1.5 GB,... would not be so much of a problem from my point of view now (for the 10.000 duplicate, 5 Thumbnails use case).
And e.g. 1000 duplicates = 150 MB,.. that's nothing.
As mentioned,..only an estimation,.. you could try to load 10.000 images into the table.

Maltragor:
> Can we make this a setting?

It's your program. You don't need to ask ;-)
And of course If the dynamic loading works as expected and reduce the memory usage it could make sense to have the configuration option. Maybe with an automatic level,... e.g. below an expected usage of 100 MB always use RAM and above if selected use the disk.
(Or let the OS make the job by using the paging file^^)

0x90d:
> It's your program. You don't need to ask ;-)

But its your pull request and I only merge it when its completed. I don't merge pull requests where I have to change or add things in order to work. This is just taking more of my limited time I have available to maintain this repo.

PR #169, opened 2021-08-18
Selection rules for displayed MediaInfo adapted to the default rules …

Maltragor:
As announced in  _Bugfix/wrong metadata_ #152 I changed the selection criteria.
(Bugfix: MKV file: Image resolution used instead of video resolution #137)

Because this has made the reading of the default flag superfluous and thus the original JsonReade can still be used, this  branch has nothing in common with the previous variant, hence the new PR.

Since the audio criterion is based on the channel count, I still had to make a small change to the JsonReader and MediaInfo, because "channels" were not considered so far.


0x90d:
Thank you very much!

PR #182, opened 2021-08-24
Improved (error) flag handling

Maltragor:
This PR should be seen as a basis for discussion. Details such as coding style etc. should be left out of the discussion until it has been clarified whether the goals pursued with the changes are desired.

Conclusion of the goals:
- More flexibility to control if files should be processed again
- Less unnecessary processing of files we can't read

Details:
- The ExitCode of ffmpeg and ffprobe is evaluated and used as failure criteria
- If ffprobe can't get the meta data this will now really cause that the MetadataError is set. (Based on the ExitCode and in addition based on an json object not containing at least one of the two sections. )
  - In the original version GetMediaInfo() only "failed" if a exception occurred e.g. WaitForExit() timeout.
  - In principle one could not return on GetMediaInfo() failure and go on to get the grayBytes. But as we have no Duration in this case we would only get grayBytes from frame 0. So at best several videos without metadata could still be compared "meaningfully". In this case we just would not return on error.
- InvalidEntry() now also filters MetadataError and Thumbnail Errors

As now videos flagged with MetadataError or ThumbnailError are not processed a second time, enabling e.g. incompatible hardware acceleration would result in setting the flags so further searches with disabled acceleration wouldn't process the videos again. This behaviour we also know from the ToDark flag even after the number of thumbnails was changed.
In addition I assume that adding the ManuallyExcluded flag could not be removed.
=> So I added two database actions allowing to remove all the error flags and to remove the ManuallyExcluded from all database entries. In addition when disabling cuda acceleration all ThumbnailError flag are removed automaticity and when changing the number of Thumbnails the ToDark flag is removed. This is done delayed and not while changing the Settings.

(As this branch is based on your master the extended ffmpeg logging is not considered at the moment. So there are only some t.b.d. comments where the additional ExitCode,... should be logged)

### DIFF
--- a/VDF.Core/EntryFlags.cs
+++ b/VDF.Core/EntryFlags.cs
@@ -26,9 +26,9 @@ namespace VDF.Core {
 		MetadataError = 8,
 		TooDark = 16,
 
-		AllErrors = ThumbnailError | MetadataError | TooDark
+		AllErrors = ThumbnailError | MetadataError | TooDark,
+		RemovableFlags = ManuallyExcluded| ThumbnailError | MetadataError | TooDark
 	}
-
 	public static class EntryFlagExtensions {
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public static bool Any(this EntryFlags f, EntryFlags checkFlags) => (f & checkFlags) > 0;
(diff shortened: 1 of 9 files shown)


0x90d:
This PR addresses too many different changes. It would be nice if you could create several PRs with each of these changes so its not only easier for reviewing but I can also merge changes I'm okay with.

PR #207, opened 2021-09-22
probeing different ffmpeg library paths

Maltragor:
("avformat",.. are elements of FFmpeg.AutoGen FFmpeg.cs LibraryDependenciesMap)
avformat depends on avcodec and avutil; avformat, avcodec and swscale are the libraries needed by avformat_alloc_context(), sws_getContext(),.. but it is not impossible that I have overlooked something else. But as the original implementation only checked avcodec it should be no problem. (Alternatively you could use "avdevice" this lib depends on everything and more)

Of course it would be possible to remove the ScanEngine.NativeFFmpegExists and use FfmpegEngine.NativeFFmpegExists but I wanted to keep the changes to a minimum.

### DIFF
--- a/VDF.Core/FFTools/FfmpegEngine.cs
+++ b/VDF.Core/FFTools/FfmpegEngine.cs
@@ -31,9 +31,24 @@ namespace VDF.Core.FFTools {
 		public static FFHardwareAccelerationMode HardwareAccelerationMode;
 		public static string CustomFFArguments = string.Empty;
 		public static bool UseNativeBinding;
+		public static bool NativeFFmpegExists = false;
 		static FfmpegEngine() {
 			FFmpegPath = FFToolsUtils.GetPath(FFToolsUtils.FFTool.FFmpeg) ?? string.Empty;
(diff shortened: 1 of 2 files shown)


Maltragor:
With this approach the FFmpeg.AutoGen functions for path probing are called independently of the the native ffmpeg configuration. So the libraries are always loaded if they exist. If you think this could be a problem, another way of course would be to try loading the libraries when the scan is started and native ffmpeg is configured.

0x90d:
Thanks for the PR, but I've chosen a different approach. Although thinking about it at this moment I believe yours would be faster in first place and mine can be used as second....hm.

Maltragor:
In case there are problems finding/loading the libraries again, you could also take a closer look at this:

> FFmpegLoader will find and load FFmpeg libaries with the FFmpeg.AutoGen bindings. It's goal is to provide a simple and customizable way of doing this, cross-platform.

https://www.nuget.org/packages/FFmpeg.Loader/

Issue #224, opened 2021-10-05
Rename when file with same name exists crashes program


floydcg:
**Environment**
 - OS: Windows

**Describe the bug**
If you attempt to rename a file, and that filename is already in the same directory, program will crash. version 2 used to display a box that would say the file is already in the directory.

PR #225, opened 2021-10-05
Fixed: Rename when file with same name exists crashes program #224

Maltragor:


0x90d:
Sorry, I haven't seen you pull request before I've fixed it myself. I do believe its better to ask for a new name instead of just showing an error.

Issue #229, opened 2021-10-08
Rename file - Exclude file extension from Prompt?


BlokeCDP:
**Environment**
Windows

**Describe the solution you'd like**
Do not show the File Extension in the prompt when renaming the file - when selecting the text you can not simply select all text and then start typing, you have to unselect the file extension - not sure what the relevance of having the file extension in the prompt is. Other potential solution is to highlight the text in the prompt but do not highlight the file extension

V2 does not show the extension so not sure if this is a bug or is expected behaviour.

(Many thanks for the tool by the way, it is awesome).

Maltragor:
When I added the InputBox in #183, having also the file extension in the prompt was intended.
For me, the extension belongs to the file name. One of the first things in a new Windows installation for me is to reconfigure the Explorer so that the file extensions are displayed :-)

At that time there was no auto selection at all. Because most of the time I only partial modify the name or add something, auto selection was also not relevant to me.

In #230 now I have added a selection start and end parameter and used it to keep the extension not selected/highlighted.
At least that was the intention. Under (my) Linux (Mint) neither the old variant with SelectAll() nor my change had worked. Without Focus() at least the selection is set correctly. But the Focus() call makes any selection disappear again.

PR #230, opened 2021-10-08
InputBox: Auto selection range can be set.

Maltragor:
Used to keep file extension not selected when used for renaming.
->  Rename file - Exclude file extension from Prompt? #229 

May not work because of avalonia issues. At least under Linux (Mint) the selection is removed as soon as the focus is set.
(Also with SelectAll() there was already the same problem before)

Not tested under Windows


0x90d:
If its bugged we shouldn't use it if we can fix this also in a non-buggy way. But thank you for the PR

PR #248, opened 2021-12-04
Added custom actions so that the user can define commands to open one or more files

Maltragor:
- It will be possible to define your own command lines for the four actions: 
  - Open
  - Open in folder
  - Open multiple 
  - Open multiple in folder

- If setting a custom open / open in folder command, the default open action is replaced.
-> This allowes to select your favourite player or file manager (e.g. with file selection support on linux)
The multi open actions are used to open whole groups or multiple selected files.
(Open multiple in folder -> some file managers like nemo are able to open multiple directories in tabs)

- There will be two new entries in the context menu to open a group (player + in folder)
  - Only visible if the corresponding multi open command is defined

- The DoubleTapped event on the thumbnails will open all selected files (if a multi open command is defined)

- Pressing enter will open the selected file(s)
  - If the thumbnail column is focused, the open command is executed
  - If the filename column is focused, the open in folder command is executed
  - In combination with the modifier keys for row selection the workflow is faster and easier (in my opinion) than always going via the context menu

- Commands:
  - Simple form: `nemo`  -> The selected files are appended
  - Using the placeholder:  `nemo %f --tabs` -> '%f' will be replaced with the selected files
  - _Exampe: Show the selected files in terminal window: `cmd /c "echo %f & pause"`_
  - If execution fails, a logging entry will be created

### DIFF
--- a/VDF.GUI/ViewModels/MainWindowVM.cs
+++ b/VDF.GUI/ViewModels/MainWindowVM.cs
@@ -35,6 +35,7 @@ using System.ComponentModel;
 using System.Text.Json;
 using System.Linq;
 using System.Collections.Specialized;
+using System.Reflection;
 
 namespace VDF.GUI.ViewModels {
 	public class MainWindowVM : ReactiveObject {
@@ -151,6 +152,33 @@ namespace VDF.GUI.ViewModels {
 			get => _DuplicatesSelectedCounter;
 			set => this.RaiseAndSetIfChanged(ref _DuplicatesSelectedCounter, value);
 		}
+		public bool MultiOpenSupported {
+			get => !String.IsNullOrEmpty(SettingsFile.Instance.CustomCommands.OpenMultiple);
+		}
+		public bool MultiOpenInFolderSupported {
+			get => !String.IsNullOrEmpty(SettingsFile.Instance.CustomCommands.OpenMultipleInFolder);
+		}
+		static List<string> _CustomCommandList = typeof(SettingsFile.CustomActionCommands).GetProperties(BindingFlags.Public|BindingFlags.Instance).Select(p => p.Name).ToList();
+		public List<string> CustomCommandList {
+			get => _CustomCommandList;
+		}
+		PropertyInfo _SelectedCustomCommand = typeof(SettingsFile.CustomActionCommands).GetProperty(_CustomCommandList[0]);
+		public string SelectedCustomCommand {
+			get => _SelectedCustomCommand.Name;
+			set {
+				_SelectedCustomCommand = typeof(SettingsFile.CustomActionCommands).GetProperty(value);
+				this.RaisePropertyChanged(nameof(SelectedCustomCommandValue));
+			}
+		}
+		public string SelectedCustomCommandValue {
+			get => (string)_SelectedCustomCommand.GetValue(SettingsFile.Instance.CustomCommands);
+			set {
+				 _SelectedCustomCommand.SetValue(SettingsFile.Instance.CustomCommands, value);
+				 this.RaisePropertyChanged(nameof(MultiOpenSupported));
+				 this.RaisePropertyChanged(nameof(MultiOpenInFolderSupported));
+			}
+		}
+		
 		public string TotalSizeRemoved => TotalSizeRemovedInternal.BytesToString();
 #if DEBUG
 		public static bool IsDebug => true;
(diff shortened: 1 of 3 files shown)


VDF.GUI/ViewModels/MainWindowVM.cs
```diff
@@ -151,6 +152,33 @@ public sealed class CheckedGroupsComparer : System.Collections.IComparer {
...
+			get => _CustomCommandList;
+		}
+		PropertyInfo _SelectedCustomCommand = typeof(SettingsFile.CustomActionCommands).GetProperty(_CustomCommandList[0]);
+		public string SelectedCustomCommand {
+			get => _SelectedCustomCommand.Name;
+			set {
+				_SelectedCustomCommand = typeof(SettingsFile.CustomActionCommands).GetProperty(value);
+				this.RaisePropertyChanged(nameof(SelectedCustomCommandValue));
+			}
+		}
```
Symbai, line 172: Avalonia is currently doing some work to make trimming and AOT possible which I would like to use then. But using Reflection is not possible in AOT.

Symbai (changes requested):
Very good job, just small code-style things I can change myself later. Only using Reflection is something I would like to avoid

Kutabarie:
Can you please explain how this would work in Windows? I still can't open more than 1 file at once.

Maltragor:
The mechanism is identical in Windows and Linux. 
Since you did not describe in which application you want to open the files, I can only describe the functionality in general.
And of course the application you have chosen must be able to open multiple files passed as command line parameters!

For demonstration purposes I have chosen cmd.exe as application to simply output the filenames in a console window: 
`cmd /c echo %f & pause`
(While executing the command `%f ` gets replaced by the selected file names)

In the following I used the `OpenMultipleInFolder` command:
Unbenannt3

The command is executed **either**:
Unbenannt
-> Select multiple rows 
-> right click on one of the selected rows 
-> click on "open in folder"

**Or**:
Unbenannt2
-> Select multiple rows 
-> the active selected cell must belong to the **path** column (image2.jpg with white rectangle) 
-> press _enter_ on the keyboard to execute the command

This will execute the configured `cmd /c echo %f & pause` command line so that the file names are displayed in a console window:
Unbenannt4

----------------
Alternatively, the configurable `OpenMultiple` command will be called when:
-> Select multiple rows 
-> the active selected cell must belong to the **thumbnail** column
-> press _enter_ on the keyboard to execute the command

Kutabarie:
Thank you for the reply. I thought by default I would be able to open all the selected files using the default media player as if I have selected the files in Windows Explorer and pressed Enter, but I guess i'll have to learn if I can work it out if i'm able to use commands on my player. What I had in mind is that if I selected 3 files and pressed enter, it would open 3 windows of MPC or PotPlayer at once.

Maltragor:
> open all the selected files using the default media player as if I have selected the files in Windows Explorer and pressed Enter

Then you can use this as command:
`cmd /c for %x in (%f) do start "" "%~x"`

This will briefly open a console window, loop over all selected file names and opens each file with the default player
`cmd /c ` -> executes the subsequent parameters in cmd.exe
`%f`        -> file name placeholder of VDF. It will be replaced by VDF before calling the command
`for %x in ([parameter list]) do` -> A loop. Executes the subsequent command for each element of `[parameter list]` (accessible by `%x`) 
`start "" "%~x" `  -> `start` is a function of cmd.exe to open a file using the default application for the corresponding file type (or to execute a process detached)

-------------
If you want to use a specific player you could use this command:
`cmd /c for %x in (%f) do start "" "C:\Program Files\MPC-HC\mpc-hc64.exe" "%~x"`

Of course you would replace `C:\Program Files\MPC-HC\mpc-hc64.exe` by your desired application.
_(But keep the quotation marks. They are necessary to support paths and filenames with spaces)_


Alternatively you can save the following as text file with filename e.g. `RunPlayer.cmd` in your VDF root directory:
_(... and of course again change the player as desired ...)_
```
@echo off

set PLAYER="C:\Program Files\MPC-HC\mpc-hc64.exe"

for %%x in (%*) do (
  start "" %PLAYER% "%%~x"
)
```
and only insert `RunPlayer.cmd %f ` as command in VDF
_Functionally it is identical to the previous variant, only divided differently, so that in VDF the whole command does not have to be entered._

Kutabarie:
Thank you so much, so pretty much completes my experience with this program!

PR #260, opened 2022-01-04
Bugfix, Fullscreen window width greater than screen width

4rch0n:
Via: https://github.com/AvaloniaUI/Avalonia/issues/5581


4rch0n:
Before change, when window is maximized (windows 10):
image
After change, when window is maximized (windows 10):
image

PR #262, opened 2022-01-06
Bugfix, busy indicator - progress bar animation not disabled when idle

4rch0n:
via https://github.com/AvaloniaUI/Avalonia/issues/2012

At this moment if you have IsIndeterminate set to true progress bar animation is constantly calculated, even if progress bar is not visible -- which causes non trivial average cpu load while application is idle.

The simplest approach to fix it in this case is to bind IsIndeterminate to the same value that controls visibility of the progress bar.

### DIFF
--- a/VDF.GUI/Views/MainWindow.xaml
+++ b/VDF.GUI/Views/MainWindow.xaml
@@ -1111,7 +1111,7 @@
                         MaxWidth="200"
                         BorderBrush="LimeGreen"
                         BorderThickness="10"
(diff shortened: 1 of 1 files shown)


4rch0n:
You have here also issue with presentation of the progress bar itself (https://github.com/4rch0n/videoduplicatefinder/commit/cfae256905b9872aba6bbbc0d6f98d152f21aa11) but let's get back to it later, for now this change is rather more important.

0x90d:
Thanks!

PR #263, opened 2022-01-07
Bugfix, progress bar calculations

4rch0n:
In gathering phase you are not incrementing progress bar for some cases which usually results in completing work when progress bar has something like 30% (because for example database is much bigger than number of scanned items narrowed down by included dirs), I have added here missing incrementations.

In comparison phase correct number of ticks for progress bar is not `n` but `(n*(n-1))/2` which is limit of a sequence
for `sum(n-k, k=1..n), n>0`, because of this change I had to change types for `scanProgressMaxValue` and `processedFiles` from `int` to `long`, to ensure that we are in bounds of the numeric type (progress bars in avalonia are using double for `value` and `max` properties so long will work here without issues, I assumed double would be overkill)

This bugfix fixes also calculation of remaining time, which at this moment is constantly increasing due to incorrect number of ticks.


4rch0n:
On second thought in comparison phase we can also move IncrementProgress to outside loop, that will work even better :), PR updated.

PR #269, opened 2022-01-10
Disable default re-scan of failed entries in consecutive scans; Add setting to enable it if needed;

4rch0n:
At this moment each time scan is run engine will try to sample each previously failed file - for most cases in regular flow situation won't change, it will fail again -- you would have to update ffmpeg in the meantime or other dependencies, so forcing these scans each time is not bringing much value.

In this PR i have added `Always retry failed sampling` option which `If activated scan engine will try to generate thumbprints of failed entries each time it is run. Otherwise consecutive attempt will be taken only after cleanup of the database.`

Something to consider :)

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -221,7 +221,12 @@ namespace VDF.Core {
 					while (pauseTokenSource.IsPaused) Thread.Sleep(50);
 
 					entry.invalid = InvalidEntry(entry);
-					if (entry.invalid) {
+
+					bool skipEntry = false;
+					skipEntry |= entry.invalid;
+					skipEntry |= entry.Flags.Has(EntryFlags.ThumbnailError) && !Settings.AlwaysRetryFailedSampling;
+
+					if (skipEntry) {
 						IncrementProgress(entry.Path);
 						return ValueTask.CompletedTask;
 					}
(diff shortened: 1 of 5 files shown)


0x90d:
This is by design. If the user changes certain settings the result may (or may not) change. Its not only when updated ffmpeg or ffprobe. Your setting makes sense. I just ask myself if it should automatically be off when a user changes the previously mentioned settings. For example ignore black/white pixels. Also I want this setting to be `TRUE` by default. Otherwise I fear I receive too many false positive issues here of people complaining results appear / disappear. Which is already hard enough to track down.

4rch0n:
Hm ok, but this is change for `GatherInfos` not for `CheckIfDuplicate`, options `ignoreBlackPixels || ignoreWhitePixels` are not affecting content of `videoFile.grayBytes`, they are only affect picked method of comparison, so i'm not sure why ignoring entries with `ThumbnailError` could bring unexpected results. Only custom ffmpeg arguments could matter here - but then user need to purge database to get changed effects anyway, because retry is done only for previously failed entries, not for all of them.

Right now `EntryFlags.ThumbnailError` can be set only if ffmpeg failed to return thumbnail for `grayBytes`, and then you can either always try to get it each scan like right now or just ignore failed entries (which has been added here) until user try to cleanup database which will remove entries with failed thumbnails via already existing function `Database.RemoveWhere(a => !File.Exists(a.Path) || a.Flags.Any(EntryFlags.MetadataError | EntryFlags.ThumbnailError));`

PR #275, opened 2022-01-13
Catch (and ignore) spurious System.IO.DirectoryNotFoundException when traversing network shares

willfe:
When running Video Duplicate Finder over a group of paths that includes one or more network shares, occasionally a path on the share is inaccessible (permission denied, temporary error on the server, etc.) and this rather unhelpful exception gets thrown as a result (along with the equally unhelpful message `Could not find a part of the path 'blah'`). It's harmless (the right behavior is to simply ignore that directory and its children and move on), but the program crashes because the exception doesn't get handled.

This PR takes care of that. I've been running the program with this change in place for a few days now without incident. Performance and functionality are unchanged, and now it won't crash when a network share acts grumpy.

### DIFF
--- a/VDF.Core/Utils/FileUtils.cs
+++ b/VDF.Core/Utils/FileUtils.cs
@@ -56,9 +56,15 @@ namespace VDF.Core.Utils {
 			if (ignoreHardLinks)
 				enumerationOptions.AttributesToSkip |= FileAttributes.ReparsePoint;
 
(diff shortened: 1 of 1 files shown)


Stevecaboose:
I have a change coming that rewrites this whole method. I can add this

0x90d:
Added this via latest commit, thanks for pointing this out

PR #280, opened 2022-01-16
Rescan action

4rch0n:
In case user would like to quickly recompare all results gathered by scan action with different comparison parameters (like percent threshold etc) it is possible with this new action to skip gathering phase and just redo comparison, should increase convenience of finding sweet spots for particular collections. Of course regular full scan needs to be done at least once for this new button to be enabled.

image

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -102,8 +102,11 @@ namespace VDF.Core {
 
 		public async void StartCompare() {
 			Logger.Instance.Info("Scan for duplicates...");
-			if (!cancelationTokenSource.IsCancellationRequested)
+			if (!cancelationTokenSource.IsCancellationRequested) {
+				cancelationTokenSource.Cancel();
+				cancelationTokenSource = new CancellationTokenSource();
 				await Task.Run(ScanForDuplicates, cancelationTokenSource.Token);
+			}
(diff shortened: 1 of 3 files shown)


4rch0n:
PR updated due to bug in initial version showing non existing -1B entries on rescan after deleting some duplicates.

4rch0n:
Hm, this PR was created too early, I can see at least two things that needs to be fixed, for now I will withdraw it and recreate it when it's in better shape.

4rch0n:
Minimum things that need to be fixed:
- handling timers on rescan
- after changing number of thumbnails in settings rescan should be disabled until next full scan, right now rescan is ignoring this change and is comparing with value of this setting set on full scan

Anyway, please let me know if this feature is something that makes sense to you, I will improve it anyway for myself and my fork, but I would know if it makes sense to recreate this particular PR here in the future.

0x90d:
If I understand you correct, this PR would skip file enumeration, getting gray bytes, etc. Just taking everything which is in database and checking for duplicates (with the new settings), right?

If so then I'm very interested. Thanks for your past/current and upcoming work on this project btw ❤

4rch0n:
Yup, that was the idea, Just added a bit more polished version.
Thanks, it's really nice project, so I'm glad I can help a bit here and there.

PR #284, opened 2022-01-22
Add Rescan action

4rch0n:
Feature initially previewed in https://github.com/0x90d/videoduplicatefinder/pull/280

> In case user would like to quickly recompare all results gathered by scan action with different comparison parameters (like percent threshold etc) it is possible with this new action to skip gathering phase and just redo comparison, should increase convenience of finding sweet spots for particular collections. Of course regular full scan needs to be done at least once for this new button to be enabled.

image

In the meantime I have fixed few things:
- On change of thumbnail setting - user is informed that full scan will be required, rescan is disabled
- On remove - DB is correctly updated and removed items won't be shown on rescan
- On rename - DB is correctly updated and renamed items will be shown correctly on rescan
- Timers are handled correctly in both scan and rescan actions
- Minor refactoring of `Prepare` function to reflect split of the search action

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -35,6 +35,7 @@ namespace VDF.Core {
 		public event EventHandler<ScanProgressChangedEventArgs>? Progress;
 		public event EventHandler? BuildingHashesDone;
 		public event EventHandler? ScanDone;
+		public event EventHandler? ScanAborted;
 		public event EventHandler? ThumbnailsRetrieved;
 		public event EventHandler? FilesEnumerated;
 		public event EventHandler? DatabaseCleaned;
@@ -83,7 +84,7 @@ namespace VDF.Core {
 		public static bool NativeFFmpegExists => FFTools.FFmpegNative.FFmpegHelper.DoFFmpegLibraryFilesExist;
 
 		public async void StartSearch() {
-			Prepare();
+			PrepareSearch();
 			SearchTimer.Start();
 			ElapsedTimer.Start();
 			Logger.Instance.InsertSeparator('-');
@@ -97,6 +98,19 @@ namespace VDF.Core {
 			Logger.Instance.Info($"Finished gathering and hashing in {SearchTimer.StopGetElapsedAndRestart()}");
 			BuildingHashesDone?.Invoke(this, new EventArgs());
 			DatabaseUtils.SaveDatabase();
+			if (!cancelationTokenSource.IsCancellationRequested) {
+				StartCompare();
+			} else {
+				ScanAborted?.Invoke(this, new EventArgs());
+				Logger.Instance.Info("Scan aborted.");
+				isScanning = false;
+			}
+		}
+
+		public async void StartCompare() {
+			PrepareCompare();
+			SearchTimer.Start();
+			ElapsedTimer.Start();
 			Logger.Instance.Info("Scan for duplicates...");
 			if (!cancelationTokenSource.IsCancellationRequested)
 				await Task.Run(ScanForDuplicates, cancelationTokenSource.Token);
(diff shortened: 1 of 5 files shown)


4rch0n:
One more adjustment, just realized that rename and move are handled here in two different places.
- On move - DB is correctly updated and moved items will be shown correctly on rescan

4rch0n:
Sorry but one more bugfix was required, at this point I believe I have addressed everything I could find:
- Now stop button works correctly in both cases, when requested during comparison phase and when requested during gathering phase.

VDF.Core/ScanEngine.cs
```diff
@@ -211,7 +250,15 @@ public sealed class ScanEngine {
...
-
+		public static void RemoveFromDatabase(FileEntry dbEntry) => DatabaseUtils.Database.Remove(dbEntry);
+		public static void UpdateFilePathInDatabase(String newPath, FileEntry dbEntry) {
+			DatabaseUtils.Database.Remove(dbEntry);
+			dbEntry.Path = newPath;
+			DatabaseUtils.Database.Add(dbEntry);
+		}
+		#pragma warning disable CS8601 // Possible null reference assignment
+		public static Boolean GetFromDatabase(String path, out FileEntry dbEntry) => DatabaseUtils.Database.TryGetValue(new FileEntry(path), out dbEntry);
+		#pragma warning restore CS8601 // Possible null reference assignment
```
0x90d, line 261: This is only used for updating file path. I'm not feeling well with a method which first gets the item. Then calling another method which removes this item, setting a new file path to this item only to re-add this item. Why don't we just have one method calling to update the file path? And this one just updates the item without removing it?

0x90d (changes requested):

VDF.Core/ScanEngine.cs
4rch0n: Path is major part of `FileEntry` hashcode, i was pretty sure that to keep everything consistent it is required to re-add object with changed hashcode to the `HashSet` which is here implementation of the Database, but i will double check it then.
Also probably i should move this function to DatabaseUtils, not sure why I have not done this in the first place.

Regarding types i will adjust :) thanks for pointing this out.

4rch0n: Ok, so i've just squashed all commits in these PR, and adjusted `String` and `Boolean` as requested.
I have also moved UpdateFilePathInDatabase body to DatabaseUtils.
Though regarding logic of this function from what I have checked it must stays that way, or whole Database needs to be redesigned, here is simplified example that should visualize what's the issue:

```
using System;
using System.Collections.Generic;

public class Program
{
	public static void Main()
	{
		FileEntry a1 = new FileEntry("a1");
		FileEntry a2 = new FileEntry("a2");
		HashSet<FileEntry> Database = new HashSet<FileEntry>();
		Database.Add(a1);
		Database.Add(a2);
		Database.Add(a1);

		Console.WriteLine(Database.Count); // Expected output: "2" because we are trying to add the same item twice

		a1.Path = "b1"; // let's do regular update that should work with for example list

		Database.TryGetValue(new FileEntry("b1"), out FileEntry b1);
		Console.WriteLine($"b1='{b1}'");  //Expected output: "_b1=''"  because hashcode of b1 is suggesting to use hash bucket in which hashset can't find adjusted a1 object

		Database.Add(a1);
		Console.WriteLine(Database.Count); // Expected output "3" because hashcode of a1 has changed

		Database.TryGetValue(new FileEntry("b1"), out b1);
		Console.WriteLine($"b1='{b1}'");  //Expected output: _b1='b1', regular case, getting objected added in line 22. It was saved under correct hash bucket and hashset could find it with equals function

		Database.TryGetValue(new FileEntry("a1"), out a1);
		Console.WriteLine($"a1='{a1}'");  //Expected output: "a1=''" because hashcode of a1 is suggesting to use hash bucket in which 'equals' function can't find the object - in result object is not accesible by get function, but not removed
	}
	public class FileEntry
	{
		public string Path { get; set; }

		public FileEntry(string path)
		{
			this.Path = path;
		}
		public override string ToString() => Path;
		public override int GetHashCode() => StringComparer.OrdinalIgnoreCase.GetHashCode(Path);
		public override bool Equals(object obj) => obj is FileEntry entry && Path.Equals(entry.Path, StringComparison.OrdinalIgnoreCase);
	}
}
```

As you can see, as long path is part of the hashcode and we are using HashSet not much can be done to avoid re-adding item after such change.

PR #290, opened 2022-01-30
Bugfix, FileEntry folder not updated on path update

4rch0n:
Folder property at this moment is not updated while items are moved.
This PR is fixing this issue to prevent unexpected behaviors like re-appearing after being moved outside of scope set by includes list. Issue was a bit hard to spot, sorry for introducing it in the first place.


0x90d:
Not only when items are moved but also when Path is changed by user in database editor. Good catch!

PR #294, opened 2022-02-01
Fix bug where "none" hardware acceleration mode is not selected initially

### LINKED ISSUE #292: The "Hardware Acceleration" combo box has no text if "none" is the initial value
**Environment**
 - OS: Windows 11

**Describe the bug**
If the `HardwareAccelerationMode` value from the settings JSON is 0, when the app is loaded, the combo box dropdown is empty instead of showing its textual value ("none").

If it's any non-zero value, it works as expected.

**Expected behavior**
All values should show its value properly.

VDF

CrendKing:
Fix #292.

As I mentioned, the custom command combo box is fine but the HA mode is not. So tried a few things to see where the difference is. Turns out, it is the type. Only enumerable of string works.

Once switch to string, it's working again. Also according to [this answer](https://stackoverflow.com/a/55624925/153133) `ConvertAll` is better than `Select`, so I used the former.

I'm not sure of those `RaisePropertyChanged` parts. Correct me if wrong.

### DIFF
--- a/VDF.GUI/ViewModels/MainWindowVM.cs
+++ b/VDF.GUI/ViewModels/MainWindowVM.cs
@@ -92,10 +92,18 @@ namespace VDF.GUI.ViewModels {
 			set => this.RaiseAndSetIfChanged(ref _IsPaused, value);
 		}
 #pragma warning disable CA1822 // Mark members as static => It's used by Avalonia binding
-		public IEnumerable<Core.FFTools.FFHardwareAccelerationMode> HardwareAccelerationModes =>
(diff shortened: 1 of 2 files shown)


0x90d:
I appreciate your attempt to fix this "issue" but its just a workarond of an Avalonia bug and I don't really want to merge this one. Especially because its just a minor and visual only. If you like you can report it on their repo.

CrendKing:
Understandable. If Avalonia's combo box is supposed to support non-string items, it's certainly their issue. Feel free to close this.

PR #340, opened 2022-08-22
Add Select All Except Oldest selection to GUI

Nighteyez07:
Add a new menu item to the Selection capability that selects all items in a group except for the oldest determined by DateCreated property on the file. This will be the first of a few additional selection items if this PR is accepted. While there is the custom expression capability, it has very limited functionality. This PR aims to bring an reasonable method of selecting duplicates for a group.

### DIFF
--- a/VDF.GUI/ViewModels/MainWindowVM.cs
+++ b/VDF.GUI/ViewModels/MainWindowVM.cs
@@ -1100,6 +1100,25 @@ namespace VDF.GUI.ViewModels {
 			}
 
 		});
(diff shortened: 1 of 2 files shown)


chchia:
it might be good to also add to allow user select all except newest. (opposite way)

0x90d:
The reason I have not reviewed / merged this PR is because I believe the whole selection feature needs some redesign. I don't want to add tons of individual selections to the GUI. Instead it should be possible for others to build their very own selections the want. For this I've added the custom selection feature but it doesn't support all cases and it might be not usable for some people due to missing knowledge of basic programming language.

Issue #295, opened 2022-02-06
Docker Container


dsrtusr88:
**Environment**
 - OS: Docker

**Describe the solution you'd like**
Has anyone created a stand-alone container for use in docker?

piotrminkina:
I started my own project today https://github.com/piotrminkina/docker-videoduplicatefinder. I didn't know someone had already started work on it. If you still want to give it a try, run a container with bellow command.

```shell
docker volume create vdf_config
docker run \
  --name vdf \
  --rm \
  -e PUID=1000 \
  -e PGID=1000 \
  -e UMASK=002 \
  -e KEYBOARD=en-us-qwerty \
  -e TZ=Europe/Warsaw \
  -v vdf_config:/config/ \
  -v ~/Photos/:/Photos/ \
  -p 3000:3000 \
  piotrminkina/videoduplicatefinder:3.0.x
```

My project is closely based on https://hub.docker.com/r/linuxserver/digikam.

0x90d:
There is also https://github.com/jlesage/docker-video-duplicate-finder

PR #347, opened 2022-10-18
Docker

kozobot:
Adding a docker build for this project. 

The end result is a Docker image that can be used to run VDF in a container and is accessible via a browser based VNC client.

### DIFF
--- a/Docker/VDF.desktop
+++ b/Docker/VDF.desktop
@@ -1,0 +1,5 @@
+[Desktop Entry]
+Type=Application
+Name=VDF
+Exec=/root/Desktop/VDF/VDF.GUI
+Icon=gnome-settings-theme
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,11 @@ RUN apt-get update \
         dotnet-runtime-6.0 \
     && rm -rf /var/lib/apt/lists/*
 
+# Copy the built files to the Desktop
 WORKDIR /root/Desktop
 COPY --from=build-env /app/VDF.GUI/out ./VDF
+
+# Copy the shortcut to the desktop
+COPY Docker/VDF.desktop .
 
 ENTRYPOINT ["/startup.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,48 @@
+### Build the VDF.Core
+FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build-env
+
+# Copy core as a layer
+WORKDIR /app/VDF.Core
+COPY VDF.Core ./
+    
+# Copy GUI as a layer and run the restore
+WORKDIR /app/VDF.GUI
+COPY VDF.GUI ./
+RUN dotnet restore
+    
+# Build the projects
+RUN dotnet publish -c Release -o out
+
+### Copy the compile app to the runtime location
+# Build runtime image
+FROM mcr.microsoft.com/dotnet/aspnet:6.0
+WORKDIR /app
+COPY --from=build-env /app/VDF.GUI/out .
+#ENTRYPOINT ["dotnet", "VDF.GUI.dll"]
+
+### Set up the VNC/Web environment
+FROM dorowu/ubuntu-desktop-lxde-vnc
+
+# Install .NET dependencies
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4EB27DB2A3B88B8B
+RUN apt-get update \
+    && apt-get install -y libx11-dev \
+    && apt-get autoclean -y \
+    && apt-get autoremove -y \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install .NET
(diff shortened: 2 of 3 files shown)


Docker/VDF.desktop
```diff
@@ -0,0 +1,5 @@
+[Desktop Entry]
```
kozobot, line 1: This is the file that gets added to the container's desktop to easily run VDF.

The user will need to click "Execute" after clicking on it.  Not sure of all the syntax this type of shortcut supports.  I assume it would be possible to default that.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+    
+# Build the projects
+RUN dotnet publish -c Release -o out
+
+### Copy the compile app to the runtime location
+# Build runtime image
+FROM mcr.microsoft.com/dotnet/aspnet:6.0
+WORKDIR /app
+COPY --from=build-env /app/VDF.GUI/out .
+#ENTRYPOINT ["dotnet", "VDF.GUI.dll"]
```
kozobot, line 21: This line should probably be deleted.  We don't want VDF as the endpoint since there wouldn't really be a way to access the UI.  That is where VNC comes in.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+### Set up the VNC/Web environment
+FROM dorowu/ubuntu-desktop-lxde-vnc
+
```
kozobot, line 25: this is a docker image i found that runs a 'light' version of ubuntu with the LXDE window manager and already has VNC set up.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+    && dpkg -i packages-microsoft-prod.deb \
+    && rm packages-microsoft-prod.deb \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        dotnet-runtime-6.0 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Copy the built files to the Desktop
+WORKDIR /root/Desktop
+COPY --from=build-env /app/VDF.GUI/out ./VDF
```
kozobot, line 47: This is where the built app get copied from the docker layer into the desktop and is made accessible.

README.md
```diff
@@ -40,12 +40,48 @@ Install ffmpeg / ffprobe using homebrew
...
+    --mount type=bind,source="FILE_PATH_1",target="/media/FILE_PATH_1" \
+    --mount type=bind,source="FILE_PATH_2",target="/media/FILE_PATH_2" \
+    vdf:latest
+```
+
+Access the instance via your browser at http://localhost:6080 (or whatever you set the WEB_PORT to).  On the desktop there will be a shortcut to run VDF and a VDF directory where VDF is installed to.
+
+Your files to scan will be located on /media when configuring VDF.  You must have at least one "--mount" line when starting docker, but you can have as many as you like.
+
+NOTE: It is recommended export your settings and database to your host file system.  If you remove this container these will be lost.
```
kozobot, line 62: This is really suboptimal.  It would be great if these files could be configured to be stored outside the install path.  Then we could just mount a host path into the container and always use the persistent versions of the files.

It shouldn't be too much of a problem unless people are using the full docker run command to start a new container each time.

kozobot: Maybe a command line argument to VDF.GUI that allows you to specify the location?

0x90d:
Is this file only available when people download the code? So as an end user without developing experience I cannot simply download a "docker" ready-to-go release?

MichelZ:
Closes #295

dsrtusr88:
`[~] # docker pull ghcr.io/0x90d/videoduplicatefinder
Using default tag: latest
Error response from daemon: Head "https://ghcr.io/v2/0x90d/videoduplicatefinder/                  manifests/latest": denied
`

Is there a better way to do this?

0x90d:
> it might be better coming from the maintainer.

I have zero experience with dockers at all, so I cannot help with that. And I think docker support requests are coming from end users expecting a "ready to go" release. So ideally a docker PR would take care of all of that, plus its getting reviewed from people who have docker experience.

dsrtusr88:
@kozobot how do you actually pull this docker image to use?

Gibby:
If this gets merged, I can submit a PR that will use GitHub Actions to create a docker image under the packages of the repo. 

To pull the image then would be something like 

`docker pull ghcr.io/0x90d/videoduplicatefinder`

dsrtusr88:
@0x90d can this be merged?

0x90d:
VDF uses .NET 7 now, so @kozobot would have to update this PR. But there are also some other projects like https://github.com/jlesage/docker-video-duplicate-finder which seem to be more powerful and very well documented.

0x90d:
Closing this as there is no activity. @kozobot feel free to open an updated PR if you like. For everyone else just looking for docker support I recommend checking out the repo in my post above.

PR #379, opened 2023-01-26
Check for failure when saving settings to file.

jlesage:


0x90d:
Thanks, in that case could you please check for failure on loading settings too?

jlesage:
Done!

PR #386, opened 2023-01-31
Implemented Hard Links check when reparse point option is set to ignore

David-Maisonave:
Implemented hard links check when the "Ignore Reparse Point" option is set. The same implementation includes code to show results reflecting "Ignore Reparse Point" after it's set, and without having to clear the database.


.gitignore
```diff
@@ -258,4 +258,5 @@ paket-files/
 
 # Python Tools for Visual Studio (PTVS)
 __pycache__/
-*.pyc
\ No newline at end of file
+*.pyc
+VDF.GUI/Properties/launchSettings.json
```
0x90d, line 262: This is an unwanted change

VDF.GUI/VDF.GUI.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ApplicationIcon>Assets\icon.ico</ApplicationIcon>
+    <StartupObject>VDF.GUI.Program</StartupObject>
```
0x90d, line 7: This is also unwanted

VDF.Core/ScanEngine.cs
```diff
@@ -684,5 +706,41 @@ public sealed class ScanEngine {
...
+		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+		static extern IntPtr FindFirstFileNameW(string lpFileName, uint dwFlags, ref uint StringLength, StringBuilder LinkName);
+		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+		static extern bool FindNextFileNameW(IntPtr hFindStream, ref uint StringLength, StringBuilder LinkName);
+		[DllImport("kernel32.dll", SetLastError = true)]
+		static extern bool FindClose(IntPtr hFindFile);
+		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+		static extern bool GetVolumePathName(string lpszFileName, [Out] StringBuilder lpszVolumePathName, uint cchBufferLength);
+		public static readonly IntPtr INVALID_HANDLE_VALUE = (IntPtr)(-1); // 0xffffffff;
+		public const int MAX_PATH = 65535; // Max. NTFS path length.
```
0x90d, line 721: These shouldn't be public

0x90d (commented):
I'm not fully convinced by this implementation. As far as I know FAT32 doesn't support hard links. How does the implementation behave on files which are on a FAT32 volume? Or on other volumes than NFTS ? This is one of the reasons why I want to rely on .NET APIs most of the time.

VDF.GUI/VDF.GUI.csproj
David-Maisonave: Removed

.gitignore
David-Maisonave: Removed.
That was my bad attempt to not include changes to launchSettings.json in this update.

But I have a question on this issue.  I could not get Visual Studio 2022 to start VDF in debug mode without adding the working directory in the launchSettings.json file. How are you able to start the program in Visual Studio without specifying a working directory?

VDF.Core/ScanEngine.cs
```diff
@@ -684,5 +706,41 @@ public sealed class ScanEngine {
 				cancelationTokenSource.Cancel();
 		}
 	}
+
+	public static class HardLinkHelper {
+		#region WinAPI P/Invoke declarations
+		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+		static extern IntPtr FindFirstFileNameW(string lpFileName, uint dwFlags, ref uint StringLength, StringBuilder LinkName);
```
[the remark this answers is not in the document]
David-Maisonave: Removed StringBuilder from this class.

VDF.Core/ScanEngine.cs
David-Maisonave: Made them private.

.gitignore
David-Maisonave: reversed change.

VDF.GUI/VDF.GUI.csproj
David-Maisonave: reversed change.

David-Maisonave:
> As far as I know FAT32 doesn't support hard links. How does the implementation behave on files which are on a FAT32 volume?

If a user has a FAT32 drive there would be no reason for them to select "Exclude Reparse Point".
The only time where a user would make this selection on FAT32 is if they're scanning both an NTFS drive and a FAT32 drive on the same scan. Which is why I tested it on a FAT32 drive, and it had no impact.

IMHO, I think there should be a separate option for excluding hard links, and looking at the code, it looks like at one point there was an option for hard links. In SettingsFile.cs it has the following:
```
		[Obsolete]
		[JsonPropertyName("IgnoreHardlinks")]
		public bool IgnoreHardlinks {
			set => IgnoreReparsePoints = value;
		}
```

What do you think about adding the option back into the GUI?

0x90d:
> I think there should be a separate option for excluding hard links

I was just about to suggest it ;D Especially when the behavior / impact is unknown it should be an opt-in feature. As for the previous setting it was just mislabeled and caused confusion. The new option can be called `ExcludeHardLinks`.

VDF.GUI/VDF.GUI.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ApplicationIcon>Assets\icon.ico</ApplicationIcon>
+    <StartupObject></StartupObject>
```
0x90d, line 7: Can you remove that line please?

.gitignore
```diff
@@ -258,4 +258,3 @@ paket-files/
 
 # Python Tools for Visual Studio (PTVS)
 __pycache__/
-*.pyc
```
0x90d, line 261: Can you re-add this line please?

David-Maisonave: Sorry.  I thought that was something I added because it was highlighted in the DIFF.
re-added.

David-Maisonave:
> I was just about to suggest it ;D Especially when the behavior / impact 
I tried to add the option, and I got some errors. When it comes to GUI's I have more experience with MFC. I'll looks at it again tonight, and update the pull request.

0x90d:
I'd change the code as currently it allocates for each link `65535` chars. Its better to use a pooled array, the APIs will tell us if size is too short and how long the new size has to be. Also I think you don't have to `GetVolumePathName` API when you can also use `Path.GetPathRoot`.

Here are the changes. It also does NOT include the original file path, only the hard links.

```C#
public static partial class HardLinkHelper
{
    [LibraryImport("kernel32.dll", SetLastError = true, StringMarshalling = StringMarshalling.Utf16)]
    private static partial IntPtr FindFirstFileNameW(string lpFileName, uint dwFlags, ref int StringLength, char[] LinkName);
    [LibraryImport("kernel32.dll", SetLastError = true, StringMarshalling = StringMarshalling.Utf16)]
    [return: MarshalAs(UnmanagedType.Bool)]
    private static partial bool FindNextFileNameW(IntPtr hFindStream, ref int StringLength, char[] LinkName);

    [LibraryImport("kernel32.dll", SetLastError = true)]
    [return: MarshalAs(UnmanagedType.Bool)]
    private static partial bool FindClose(IntPtr hFindFile);


    const IntPtr INVALID_HANDLE_VALUE = -1;
    const int ERROR_MORE_DATA = 234;

    /// <summary>
    //// Returns the enumeration of hard links for the given *file* as full file paths
    /// </summary>
    public static IEnumerable<string> GetHardLinks(string filepath)
    {
        char[] buffer = ArrayPool<char>.Shared.Rent(512);
        try
        {
            string? volume = Path.GetPathRoot(filepath)?.TrimEnd('\\');
            if (volume == null)
                return Array.Empty<string>();

            List<string> links = new();
            int stringLength = buffer.Length;

            while (true)
            {
                IntPtr findHandle = FindFirstFileNameW(filepath, 0, ref stringLength, buffer);
                if (findHandle == INVALID_HANDLE_VALUE)
                {
                    if (Marshal.GetLastPInvokeError() == ERROR_MORE_DATA)
                    {
                        ArrayPool<char>.Shared.Return(buffer);
                        buffer = ArrayPool<char>.Shared.Rent(stringLength);
                        continue;
                    }
                    else
                    {
                        break;
                    }
                }
                while (true)
                {
                    bool success = FindNextFileNameW(findHandle, ref stringLength, buffer);

                    if (!success)
                    {
                        if (Marshal.GetLastPInvokeError() == ERROR_MORE_DATA)
                        {
                            ArrayPool<char>.Shared.Return(buffer);
                            buffer = ArrayPool<char>.Shared.Rent(stringLength);
                            continue;
                        }
                        else
                        {
                            FindClose(findHandle);
                            break;
                        }
                    }
                    links.Add((volume + new string(buffer)).Trim('\0'));                      
                }
                break;
            }

            return links;
        }
        catch (Exception ex)
        {
            Logger.Instance.Info(
                $"GetHardLinks: Exception, file: {filepath}, reason: {ex.Message}");
            return Array.Empty<string>();
        }
        finally
        {
            ArrayPool<char>.Shared.Return(buffer);
        }
    }
}
```

The other code can be changed to:

```C#
foreach (var link in HardLinkHelper.GetHardLinks(entry.Path)) {
	if (compItem.Path == link) {
		isDuplicate = false;
		break;
	}
}
```

David-Maisonave:
Did you change the target on the VDF project to .Net 7?
I tried your changes, but I got 38 compiler errors starting with the following error:
  - Error (active)	CS0246	The type or namespace name '**LibraryImportAttribute**' could not be found (are you missing a using directive or an assembly reference?)	VDF.Core	E:\Repos\VideoCompareProject\videoduplicatefinder\VDF.Core\ScanEngine.cs	714	

When I googled [StringMarshalling ](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.libraryimportattribute.stringmarshalling?view=net-7.0) it states it applies to .Net 7 at the bottom of the page.

0x90d:
> Did you change the target on the VDF project to .Net 7?

Not yet but I think I can do as most of the regressions in .NET 7 have been fixed. But you can also change it back to `DllImport` if you like. Changes in `GetHardLinks` function are what matters most.

VDF.GUI/Views/MainWindow.xaml
```diff
@@ -1295,6 +1295,12 @@
                                                         IsChecked="{Binding Source={x:Static Settings:SettingsFile.Instance}, Path=GeneratePreviewThumbnails}"
                                                         IsThreeState="False"
                                                         ToolTip.Tip="Deactivate to save memory by disabling preview images of duplicates" />
+                                                    <CheckBox
+                                                        VerticalAlignment="Center"
+                                                        Content="Exclude Hard Links"
```
0x90d, line 1300: Change it to "Exclude Hard Links (Windows Only)"

David-Maisonave: > Not yet but I think I can do as most of the regressions in .NET 7

Where you able to compile your changes using .NET 6?

> Change it to "Exclude Hard Links (Windows Only)". Or leave a note in the tooltip that it only works on Windows.

I'll see if I can hide this option when not running on Windows.

0x90d: Not hiding, just rename. So people know this feature exists. The code above won't compile as `LibraryImport` is a .NET 7 feature. Like I said you can change it back to `DllImport`.

David-Maisonave:
> Change it to "Exclude Hard Links (Windows Only)". Or leave a note in the tooltip that it only works on Windows.

Done

> The code above won't compile as LibraryImport is a .NET 7 feature. Like I said you can change it back to DllImport.

Sorry, I misunderstood your previous comment. However, it's not just the LibraryImport that's the issue if using .NET 6.  StringMarshalling also can't be used unless it's .NET 7.


> string? volume = Path.GetPathRoot(filepath)?.TrimEnd('\\');

**GetPathRoot** can not be used, because that will fail on a drive that is [mounted on an empty folder](https://learn.microsoft.com/en-us/windows-server/storage/disk-management/assign-a-mount-point-folder-path-to-a-drive).
Example: E:\MyVideoDrives\CartoonDrive
MountToEmptyFolderSnapshot


In this example, "E:\MyVideoDrives\CartoonDrive" is the volume path, but GetPathRoot will return "E:" instead.

In testing this out, I found there was a bug in my code as well, and I had to zero out the buffer before each call to FindFirstFileNameW and FindNextFileNameW. I just updated my branch with the fix.


> foreach (var link in HardLinkHelper.GetHardLinks(entry.Path)) {
	if (compItem.Path == link) {
		isDuplicate = false;
		break;
	}
}

The above code is less efficient then the original code, because it's always doing a string comparisons for every file.  In the most common use case, most of the files are not going to be hard links, so it's better to first check if there are multiple links before checking if any of the links match the "**compItem.Path**".
```
if (links.Count > 1)
```
A Count check takes a very small amount of computing time compare to a string comparison check.

I'm getting a reproducible exception thrown with VDF when I try to add the following path to the list of files to scan:
`E:\Temp\zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\Z_Drv_Alias\more\VideoTestFiles\SimilarFiles\RP_Test
`
This path is 264 characters long, so I think there must be some code with a 255 byte limit causing the issue. The following stack doesn't seem to show what's throwing the exception.
[StackThrowingException.txt](https://github.com/0x90d/videoduplicatefinder/files/10583314/StackThrowingException.txt)

I'll take a look

0x90d:
> The above code is less efficient then the original code, because it's always doing a string comparisons for every file

Considering the amount of pinvokes we are calling and considering the amount of memory we allocate per file already with this implementation, I think that a simple string check isn't a problem. Especially as in .NET string checks are extremely fast and very well optimized.

Besides, in my code above, if there isn't any hardlink then there isn't any string check as well. Unlike yours, mine only returns any hard links but not the original path.


> I'm getting a reproducible exception thrown with VDF when I try to add the following path to the list of files to scan

From the stacktrace the first line which is within VDF code is this one ```
/_/src/Avalonia.Controls/ApplicationLifetimes/ClassicDesktopStyleApplicationLifetime.cs(121)
VDF.GUI.dll!VDF.GUI.Program.Main(string[] args) Line 30
```. It looks like the exception is thrown when you start VDF and not when you add the path to the list. I assume you have opt in long path in Windows? Maybe Avalonia has a problem with long paths.

**Comment by David-Maisonave** [comment]
> It looks like the exception is thrown when you start VDF 

It starts up fine.  This happens when I add the long path. If you have windows, this should be easy to reproduce on your setup by creating a long path.  You can create a long path by creating a junction point on a 200 character path pointing to the base of of a path having 70 characters. Or you can mount a drive on a 200 character path.

> mine only returns any hard links but not the original path.

I didn't realize the code was doing that.  If I'm reading the code right, that's not going to work. 
The code is assuming that the first path will be equal to the input filepath. But that's a false assumption.  The first path could be equal to the filepath, or it could just as well be equal to the other hard link.

I verified this by with the below code where I added two static variables called **FirstPassCount** and **NextPassCount**. The added code just counts how many times filepath is equal to the value fetched from FindFirstFileNameW and how many times it was not by incrementing NextPassCount.
```
	public static class HardLinkHelper {
		#region WinAPI P/Invoke declarations
		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
		static extern IntPtr FindFirstFileNameW(string lpFileName, uint dwFlags, ref uint StringLength, char[] LinkName);
		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
		static extern bool FindNextFileNameW(IntPtr hFindStream, ref uint StringLength, char[] LinkName);
		[DllImport("kernel32.dll", SetLastError = true)]
		static extern bool FindClose(IntPtr hFindFile);
		[DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
		static extern bool GetVolumePathName(string lpszFileName, [Out] char[] lpszVolumePathName, uint cchBufferLength);
		private static readonly IntPtr INVALID_HANDLE_VALUE = (IntPtr)(-1); // 0xffffffff;
		private const int MAX_PATH = 65535; // Max. NTFS path length.
		static private int FirstPassCount = 0;
		static private int NextPassCount = 0;
		#endregion
		/// <summary>
		//// Returns the enumeration of hard links for the given *file* as full file paths, which includes the input path itself.
		/// </summary>
		public static List<string> GetHardLinks(string filepath) {
			List<string> links = new List<string>();
			try {
				Char[] sbPath = new Char[MAX_PATH + 1];
				uint charCount = (uint)MAX_PATH;
				GetVolumePathName(filepath, sbPath, (uint)MAX_PATH);
				string volume = new string(sbPath).Trim('\0');
				volume = volume.Substring(0, volume.Length - 1);
				Array.Clear(sbPath, 0, MAX_PATH);
				IntPtr findHandle;
				bool FirstPass = true;
				if (INVALID_HANDLE_VALUE != (findHandle = FindFirstFileNameW(lpFileName: filepath, 0, ref charCount, sbPath))) {
					do {
						string HardLinkPath = (volume + new string(sbPath)).Trim('\0');
						links.Add(HardLinkPath); // Add the full path to the result list.
						charCount = (uint)MAX_PATH; // Prepare for the next FindNextFileNameW() call.
						Array.Clear(sbPath, 0, MAX_PATH);
						if (FirstPass) {
							if (HardLinkPath == filepath)
								FirstPassCount++;
							else
								NextPassCount++;
						}
						FirstPass = false;
					} while (FindNextFileNameW(findHandle, ref charCount, sbPath));
					FindClose(findHandle);
				}
			}
			catch (Exception ex) {
				Logger.Instance.Info(
					$"GetHardLinks: Exception, file: {filepath}, reason: {ex.Message}, stacktrace {ex.StackTrace}");
			}
			return links;
		}
```

I ran the code over a path that had 9 sets of hard links. After scanning the path, FirstPassCount = 5 and NextPassCount = 4.
That basically shows there's about a 50/50 chance that FindFirstFileNameW will return the path equal to the input filepath value.

The implementation needs to evaluate the value returned from FindFirstFileNameW and from FindNextFileNameW.

**Comment by David-Maisonave** [comment]
> const IntPtr INVALID_HANDLE_VALUE = -1;

The above line doesn't compile, and I get the following error:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error (active)	CS0283	The type 'IntPtr' cannot be declared const	VDF.Core	E:\Repos\VideoCompareProject\videoduplicatefinder\VDF.Core\ScanEngine.cs	727	
```
I recommend using the code I have in the update. It compiles on .NET 6.0, it's simple, no bug, it's been well tested on hundreds of hard links, and it's efficient.

Expanding the code increases the chances of adding bugs.

David-Maisonave:
I like your **foreach** code change, so I modified GetHardLinks so that it returns an empty list if **links.Count** is less then 2.
I added a second input variable (**ReturnEmptyListIfOnlyOne** ) so that GetHardLinks works as expected in general, and only returns an empty list if ReturnEmptyListIfOnlyOne is set to true and **links.Count** is less then 2.

0x90d:
I appreciate your effort but I don't want to merge it in its current state. Despise the false assumption, my concerns above still exist. And the second parameter doesn't really make it better (more the opposite). Since its always true, it's completely redundant.

0x90d:
I think it saves us both some time if I just add my implementation. I gave credits to you for the original idea / implementation. Again thanks for your efforts.

PR #387, opened 2023-01-31
Some exported files were not having suggested start location.

jlesage:


0x90d:
This is by design. Leaving it unset should make the file dialog using the last used location. Most of the case when you export log, database or scan results, you don't save them in the same folder as VDF, instead the Desktop, documents or a network drive are more common places.

jlesage:
Should this principle also applies to ExportScanResultsIncludingThumbnails ?

Issue #385, opened 2023-01-29
Add Identical option to Filter options


David-Maisonave:
**Environment**
 - OS: Any

**Describe the solution you'd like**
On the Filter menu, add a combo box that has the following items to select from:
 - Any
 - Identical Match
 - Different (Not Identical)

Also add option so actions do NOT apply to filtered out items.

"**Any**" would be the default value and result in no filtering. **Identical match** would only display groups having identical files (same file size, video length, resolution, format, fps, and bit rate). **Different** would only display groups that have files that are not identical.

If VDF has this option, this would be a temporary workaround for the Hard Links issue.

0x90d:
> Also add option so actions do NOT apply to filtered out items.

This isn't possible unless there is a way to obtain only filtered items (except of running same filter conditions in code again). I don't know such a way.

David-Maisonave:
> This isn't possible unless there is a way to obtain only filtered items (except of running same filter conditions in code again). I don't know such a way.

I just tested an implementation which addresses this.  I added the following code in function VDF.GUI.ViewModels.DeleteInternal, after the MessageBoxButtons.Yes check.
``

			if (!string.IsNullOrEmpty(FilterByPath) || FileType.Value != FileTypeFilter.All) {
				string FilterByPath_lwr = FilterByPath.ToLower();
				HashSet<System.Guid> IncludedGroups = new HashSet<System.Guid>();
				for (var i = Duplicates.Count - 1; i >= 0; i--) {
					if (FileType.Value != FileTypeFilter.All) {
						if (FileType.Value == FileTypeFilter.Images && !Duplicates[i].ItemInfo.IsImage)
							continue;
						if (FileType.Value == FileTypeFilter.Videos && Duplicates[i].ItemInfo.IsImage)
							continue;
					}
					if (string.IsNullOrEmpty(FilterByPath) || Duplicates[i].ItemInfo.Path.ToLower().Contains(FilterByPath_lwr))
						IncludedGroups.Add(Duplicates[i].ItemInfo.GroupId);
				}
				for (var i = Duplicates.Count - 1; i >= 0; i--) {
					if (!IncludedGroups.Contains(Duplicates[i].ItemInfo.GroupId))
						Duplicates[i].Checked = false;
				}
			}

``

0x90d:
[...]

PR #393, opened 2023-02-05
Apply filters when executing a delete

David-Maisonave:
When I first starting using VDF, I deleted 100's of video's by mistake, because I thought the deletion would only apply to the filtered viewed files. I did not see the note stating filter is only visual.

The change in this pull request is small and would help avoid end users from unintentionally deleting their files.

I know it makes it so you have two different coding sections to maintain, but that's far safer then putting user's files at risk.

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -470,17 +470,13 @@ namespace VDF.Core {
 									isDuplicate = false;
 							}
 							if (isDuplicate && Settings.IgnoreHardLinks  && CoreUtils.IsWindows && entry.FileSize == compItem.FileSize) {
-								List<string> links = HardLinkHelper.GetHardLinks(entry.Path);
-								if (links.Count > 1) {
-									foreach (var link in links) {
+								foreach (var link in HardLinkHelper.GetHardLinks(entry.Path, true))
 										if (compItem.Path == link) {
 											isDuplicate = false;
 											break;
 										}
 									}
 								}
-							}
-						}
 
 						if (isDuplicate) {
 							lock (duplicateDict) {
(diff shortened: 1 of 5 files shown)


0x90d:
First of all this PR includes code from your other PR, so I cannot merge it even if I want to. When you fork this project and make multiple changes you have to create a branch for each new PR. Second, I said I don't want. So why do you think making this PR would change my opinion? I see that users may be confused about the behavior, that is why I added the note. I DO agree that it would be cool to apply actions only on items in the list, whether filter is on or off. But I won't implement such a feature as long as it means that filter has to be run twice or I have to duplicate code.

If you are really motivated enough to see this feature in VDF. You can go to https://github.com/AvaloniaUI/Avalonia/issues and make a PR which give developers access the list of visible items.

David-Maisonave:
> Second, I said I don't want. So why do you think making this PR would change my opinion? 

You never said you didn't want to do it.  Please review your comments #385. If you would have stated you didn't want to do it, I would not have posted this.

You referred to ideally using DataGridCollectionView, which I posted a comment showing that the class does have the ability to get filter data.

> If you are really motivated enough to see this feature in VDF. You can go to https://github.com/AvaloniaUI/Avalonia/issues and make a PR which give developers access the list of visible items.

DataGridCollectionView already has this feature.
According to the [DataGridCollectionView](https://reference.avaloniaui.net/api/Avalonia.Collections/DataGridCollectionView/) online documentation, the [GetItemAt](https://reference.avaloniaui.net/api/Avalonia.Collections/DataGridCollectionView/1C60C45B) member function returns the filtered items.
- Retrieve item at the given zero-based index in this DataGridCollectionView, after the source collection is **filtered**, **sorted**, and paged. 

Also [ItemCount](https://reference.avaloniaui.net/api/Avalonia.Collections/DataGridCollectionView/3F76F288) returns the number of items after filter is applied.

Is there something I'm not seeing here? I don't understand.

PR #399, opened 2023-02-14
Fixed excluded directory not being added to the correct list.

jlesage:


VDF.GUI/ViewModels/MainWindowVM_Settings.cs
```diff
@@ -110,7 +110,7 @@ public partial class MainWindowVM : ReactiveObject {
 			if (result == null || result.Count == 0) return;
 			foreach (var item in result) {
 				if (!SettingsFile.Instance.Includes.Contains(item))
```
Maltragor, line 112: Also here in the if: Includes <-> Blacklists

jlesage: Fixed, thanks!

PR #400, opened 2023-02-17
Fixed handling of file path containing the `#` character.

### LINKED ISSUE #398: Directories containing the `#` character are not handled correctly
**Environment**
 - OS: Linux

**Describe the bug**
Adding a folder under `Search Directories` that contains the `#` character is not handled correctly:
  - Display of the folder name stops at the `#` character.
  - Scan of such folder does not work.

**Expected behavior**
VDF should properly handle directories containing a `#` character.

jlesage:
Fixes #398.

The issue is that the `Uri` object built from a FilePicker result (via `GetTryUri()`) could be incorrectly parsed when the `#` character exists in the path of the file: the part following the character is considered as a `Fragment`. 
Thus, using `LocalPath` leads to an incomplete file path.  For example, for the Uri `file:///path/to/my#folder`, `LocalPath` gives `/path/to/my`.

For more details about the Uri: https://learn.microsoft.com/en-us/dotnet/api/system.uri?view=net-7.0

Note that this behaviour is seen on Linux,  but it seems different on Windows.

### DIFF
--- a/VDF.GUI/Utils/PickerDialogUtils.cs
+++ b/VDF.GUI/Utils/PickerDialogUtils.cs
@@ -19,13 +19,37 @@ using Avalonia.Platform.Storage;
 namespace VDF.GUI.Utils {
 	internal static class PickerDialogUtils {
 
+		private static string GetLocalPath(Uri uriPath) {
+			if (uriPath.IsFile) {
+				// The `LocalPath` member of the `Uri` object we got
+				// from `TryGetUri()` might be wrong if the original
+				// path contains the '#' character, or if escape
+				// sequences (e.g. '%61') occur in the path.
+				//
+				// Thus, it's better to use the `OriginalString` member
+				// instead, which contains the correct data on both
+				// Linux and Windows.  However, the `file://` scheme
+				// might be prefixed (e.g. on Linux), so it needs to be
+				// removed if present.
+				// See https://github.com/0x90d/videoduplicatefinder/pull/400
+				string scheme = uriPath.Scheme + Uri.SchemeDelimiter;
+				if (uriPath.OriginalString.StartsWith(scheme)) {
+					return uriPath.OriginalString.Substring(scheme.Length);
+				}
+				else {
+					return uriPath.OriginalString;
+				}
+			}
+			return uriPath.LocalPath;
+		}
+
 		internal static async Task<string?> OpenFilePicker(FilePickerOpenOptions options) {
 			var paths = await ApplicationHelpers.MainWindow.StorageProvider.OpenFilePickerAsync(options);
 
 			if (paths != null &&
 				paths.Count > 0 &&
 				paths[0].TryGetUri(out Uri? uriPath))
-				return uriPath.LocalPath;
+				return GetLocalPath(uriPath);
 
 			return null;
 		}
@@ -33,7 +57,7 @@ namespace VDF.GUI.Utils {
 			var path = await ApplicationHelpers.MainWindow.StorageProvider.SaveFilePickerAsync(options);
 
 			if (path != null && path.TryGetUri(out Uri? uriPath))
-				return uriPath.LocalPath;
+				return GetLocalPath(uriPath);
 
 			return null;
 		}
@@ -46,7 +70,7 @@ namespace VDF.GUI.Utils {
 			List<string> results = new();
 			foreach (var item in paths) {
 				if (item.TryGetUri(out Uri? uriPath))
-					results.Add(uriPath.LocalPath);
+					results.Add(GetLocalPath(uriPath));
 			}
 			return results;
 		}


jlesage:
As pointed by @Maltragor:
 
> It's not just the `#` sign, by the way. If escape sequences occur in the path, they would be converted by LocalPath(). `/test %61 %23 xyz` becomes `/test a # xyz`.

0x90d:
Why not just using `OriginalString` or `AbsoluteUri`? on your example it returns the correct path.

0x90d:
I already tried updating to that preview but there is a major bug where Datagrid theme is not applied. I cannot update as long as it isn' fixed. As a workaround we should probably stick to OriginalString for now. Do you want to update your PR or should I change it?

jlesage:
I already updated this PR :)

0x90d:
> I already updated this PR :)

I don't see any changes: https://github.com/0x90d/videoduplicatefinder/pull/400/files

jlesage:
I squashed my commits, so it's normal that you see a single commit.

You should see that `uriPath.OriginalString` is used as soon as `uriPath.IsFile` is true (no more check on `Fragment`).

0x90d:
This was a misunderstanding sorry. I meant to just replace `LocalFile` with `OriginalFile` in my code and get rid of that additional function which uses substring. I don't see why the extra function is necessary.

jlesage:
If `uriPath.LocalPath` can never be false, I agree with you.  But I was not sure of that, thus the reason of the extra function.

Avalonia seems to be able to use different file pickers from the OS.  I know that with Linux, for example, some file pickers allow you to enter URLs like `smb://...` to select files/directories from a Windows share.

But my knowledge with Avalonia is very limited, so I will let you take the decision on that :)

jlesage:
> I know that with Linux, for example, some file pickers allow you to enter URLs like smb://... to select files/directories from a Windows share.

FYI, from 2 different Linux distros I tried, the file picker **does not** offer this ability.

Maltragor:
Now I'm also a little confused about what is actually meant. Therefore I try to question the last comments again a little bit to exclude misunderstandings.
Should I unintentionally cause the opposite and cause even more confusion, just ignore this comment.

> If uriPath.LocalPath can never be false, [...]

should that actually be:
If _uriPath.**IsFile**_ can never be false, [...] 


>  I meant to just replace LocalFile with OriginalString in my code and get rid of that additional function which uses substring. 

Do you mean only to change
`return uriPath.LocalPath;`
to
`return uriPath.OriginalString;`

If so, that would not work.
`uriPath.OriginalString` is more then just something like _"C:\\Videos"_ or _"/home/user1/Videos"_.
`uriPath.OriginalString` would return _"file://C:\\Videos"_ or _"file:///home/user1/Videos"_
So `Substring` is used to remove _"file://"_.
`(uriPath.Scheme + Uri.SchemeDelimiter).Length` as index is only the generic version to get the length of the String _"file://"_.

> If uriPath.IsFile

Is intended to have a fallback in case it would be possible that `OriginalString` is no file-uri.

In Avalonia 11.0.0-preview4, most `OpenFolderPickerAsync()` implementations returns `files.Select(f => new BclStorageFolder(new DirectoryInfo(f))).ToArray()`
and `BclStorageFolder.TryGetUri()` returns `uri = Path.IsPathRooted(_directoryInfo.FullName) ? new Uri(new Uri("file://"), _directoryInfo.FullName) : new Uri(_directoryInfo.FullName, UriKind.Relative)`

So at least all implementations that return an absolute path as `BclStorageFolder` have the `#` and `%` problem.
Since the `BclStorageFolder` implementation has a special handling for relative paths it would be theoretically possible that the returned `uriPath` is no file-Uri. But I don't know if a `OpenFolderPicker` can return relative paths at all.

0x90d:
> If so, that would not work.

Then I need more information about the actual problem. It was reported that adding directories with a `#` character would be cut off. I couldn't reproduce because my test folder had a white space before and behind `#` character and nothing was cut off. Now with the path in the first post in this PR I can reproduce the problem. Switching to OriginalPath fixes the problem.

Input: file:///path/to/my#folder
```
AbsolutePath: /path/to/my
AbsoluteUri: file:///path/to/my#folder
LocalPath: /path/to/my
OriginalString: file:///path/to/my#folder
PathAndQuery: /path/to/my
```

Input: `C:\Test\Folder\`
```
AbsolutePath: C:/Test/Folder/
AbsoluteUri: file:///C:/Test/Folder/
LocalPath: C:\Test\Folder\
OriginalString: C:\Test\Folder\
PathAndQuery: C:/Test/Folder/
```

These are the results i get on Windows for each input. The bottom result is correct for Windows. The top path seems to be correct for Linux. I am not an expert when it comes to Linux but if we say that `file:///path/to/my#folder` is a valid path (which I assumed as it is what we use as input), then the output is valid too when its the same. In this case switching to `OriginalString` seems like a valid solution to me.

Maltragor:
OK,... there was probably one or the other wrong assumption for everyone, which then added up ;-)

I have now also output everything again under Windows and Linux with following examples:
(This are real paths in the file system)

Windows:
```
D:\share\VDF-test\my # folder
D:\share\VDF-test\my#folder
D:\share\VDF-test\test %61 %23 xyz
```
Linux:
```
/mnt/hgfs/VDF/my # folder
/mnt/hgfs/VDF/my#folder
/mnt/hgfs/VDF/test %61 %23 xyz
```

The `BclStorageFolder` row shows the String returned from the FolderPicker.
(At least in preview4, the BclStorageFolder class has the function SaveBookmarkAsync() which only returns the private _directoryInfo.FullName. Most but not all FilePickers return a BclStorageFolder. The implementation of SaveBookmarkAsync() from other derived class seems to do more then just returning the string. _directoryInfo.FullName is the path used as base to create the Uri)

```
BclStorageFolder: /mnt/hgfs/VDF/my#folder
AbsolutePath:     /mnt/hgfs/VDF/my
AbsoluteUri:      file:///mnt/hgfs/VDF/my#folder
LocalPath:        /mnt/hgfs/VDF/my
OriginalString:   file:///mnt/hgfs/VDF/my#folder
PathAndQuery:     /mnt/hgfs/VDF/my
Scheme:           file
```
```
BclStorageFolder: D:\share\VDF-test\my#folder
AbsolutePath:     D:/share/VDF-test/my%23folder
AbsoluteUri:      file:///D:/share/VDF-test/my%23folder
LocalPath:        D:\share\VDF-test\my#folder
OriginalString:   D:\share\VDF-test\my#folder
PathAndQuery:     D:/share/VDF-test/my%23folder
Scheme:           file
```
* For (my) Linux (Mint 21.1 Vera) the FolderPicker internally gets `/mnt/hgfs/VDF/my#folder`
This is also the representation `LocalPath` would provide (apart from the cut off problem).
But `OriginalString` is `file:///mnt/hgfs/VDF/my#folder`. If you would add this to VDF includes it would not work. No files would be found. So the `file://` needs to be removed in order to get a working representation.
This is the idea behind using `Substring(...)` and the reason why the unchanged OriginalString would not work as far as I can tell so far.
* On Windows 10 the FolderPicker internally gets `D:\share\VDF-test\my#folder`
And you are right, the unchanged `OriginalString` would work as it corresponds to the real path.
_wonder now why I hadn't noticed that, ...^^_ 

---------------
So at the moment it would look to me like if `OriginalString` starts with `file://` it has to be removed and otherwise leave `OriginalString` unchanged. And maybe include the fallback that `LocalPath` is used if the `Scheme` is not `file`.

--------------------
The other examples
```
BclStorageFolder: D:\share\VDF-test\my # folder
AbsolutePath:     D:/share/VDF-test/my%20%23%20folder
AbsoluteUri:      file:///D:/share/VDF-test/my%20%23%20folder
LocalPath:        D:\share\VDF-test\my # folder
OriginalString:   D:\share\VDF-test\my # folder
PathAndQuery:     D:/share/VDF-test/my%20%23%20folder
Scheme:           file

BclStorageFolder: D:\share\VDF-test\test %61 %23 xyz
AbsolutePath:     D:/share/VDF-test/test%20a%20%2523%20xyz
AbsoluteUri:      file:///D:/share/VDF-test/test%20a%20%2523%20xyz
LocalPath:        D:\share\VDF-test\test a %23 xyz
OriginalString:   D:\share\VDF-test\test %61 %23 xyz
PathAndQuery:     D:/share/VDF-test/test%20a%20%2523%20xyz
Scheme:           file

BclStorageFolder: /mnt/hgfs/VDF/my # folder
AbsolutePath:     /mnt/hgfs/VDF/my%20
AbsoluteUri:      file:///mnt/hgfs/VDF/my%20#%20folder
LocalPath:        /mnt/hgfs/VDF/my 
OriginalString:   file:///mnt/hgfs/VDF/my # folder
PathAndQuery:     /mnt/hgfs/VDF/my%20
Scheme:           file

BclStorageFolder: /mnt/hgfs/VDF/test %61 %23 xyz
AbsolutePath:     /mnt/hgfs/VDF/test%20a%20%23%20xyz
AbsoluteUri:      file:///mnt/hgfs/VDF/test%20a%20%23%20xyz
LocalPath:        /mnt/hgfs/VDF/test a # xyz
OriginalString:   file:///mnt/hgfs/VDF/test %61 %23 xyz
PathAndQuery:     /mnt/hgfs/VDF/test%20a%20%23%20xyz
Scheme:           file
```
----------------------
The modified `OpenDialogPicker` used to generate the output:
```C#
internal static async Task<List<string>?> OpenDialogPicker(FolderPickerOpenOptions options) {
	var paths = await ApplicationHelpers.MainWindow.StorageProvider.OpenFolderPickerAsync(options);

	if (paths == null || paths.Count == 0)
		return null;

	List<string> results = new();
	System.Diagnostics.Trace.WriteLine("---- OpenDialogPicker ----");
	foreach (var item in paths) {
		if (item.TryGetUri(out Uri? uriPath)) {
			System.Diagnostics.Trace.WriteLine("");
			if (item.GetType() == typeof(BclStorageFolder)) {
				// BclStorageFolder.SaveBookmarkAsync() in preview4 is just "return Task.FromResult<string?>(_directoryInfo.FullName);"
				// And _directoryInfo.FullName is used as base to create the Uri in BclStorageFolder.TryGetUri()
				var path = await item.SaveBookmarkAsync();
				System.Diagnostics.Trace.WriteLine("BclStorageFolder: " + path);
			}
			System.Diagnostics.Trace.WriteLine("AbsolutePath:     " + uriPath.AbsolutePath);
			System.Diagnostics.Trace.WriteLine("AbsoluteUri:      " + uriPath.AbsoluteUri);
			System.Diagnostics.Trace.WriteLine("LocalPath:        " + uriPath.LocalPath);
			System.Diagnostics.Trace.WriteLine("OriginalString:   " + uriPath.OriginalString);
			System.Diagnostics.Trace.WriteLine("PathAndQuery:     " + uriPath.PathAndQuery);
			System.Diagnostics.Trace.WriteLine("Scheme:           " + uriPath.Scheme);

			//results.Add(uriPath.LocalPath);
		}
	}
	return results;
}
```

jlesage:
Thanks for the tests!

> So at the moment it would look to me like if `OriginalString` starts with `file://` it has to be removed and otherwise leave `OriginalString` unchanged. And maybe include the fallback that `LocalPath` is used if the `Scheme` is not file.

This is also my conclusion.

@0x90d, do you agree ?  If yes I can adjust the PR to reflect that.

0x90d:
I honestly don't understand the choice for `Uri` and seeing the results its even more odd leaving all the problems to the developers. It should just return a `string` with full path to the object, like in every other framework. But it is what it is. Thank you @Maltragor 

@jlesage  Go ahead thanks!

jlesage:
PR updated!

VDF.GUI/Utils/PickerDialogUtils.cs
```diff
@@ -19,21 +19,46 @@
 namespace VDF.GUI.Utils {
 	internal static class PickerDialogUtils {
 
+		private static string GetLocalPath(Uri uriPath) {
+			string localPath = uriPath.LocalPath;
```
0x90d, line 23: You can remove that line

VDF.GUI/Utils/PickerDialogUtils.cs
```diff
@@ -19,21 +19,46 @@
...
+				// Linux and Windows.  However, the `file://` scheme
+				// might be prefixed (e.g. on Linux), so it needs to be
+				// removed if present.
+				// See https://github.com/0x90d/videoduplicatefinder/pull/400
+				string scheme = uriPath.Scheme + Uri.SchemeDelimiter;
+				if (uriPath.OriginalString.StartsWith(scheme)) {
+					localPath = uriPath.OriginalString.Substring(scheme.Length);
+				}
+				else {
+					localPath = uriPath.OriginalString;
```
0x90d, line 41: And return here instead

VDF.GUI/Utils/PickerDialogUtils.cs
```diff
@@ -19,21 +19,46 @@
...
+				// path contains the '#' character, or if escape
+				// sequences (e.g. '%61') occur in the path.
+				//
+				// Thus, it's better to use the `OriginalString` member
+				// instead, which contains the correct data on both
+				// Linux and Windows.  However, the `file://` scheme
+				// might be prefixed (e.g. on Linux), so it needs to be
+				// removed if present.
+				// See https://github.com/0x90d/videoduplicatefinder/pull/400
+				string scheme = uriPath.Scheme + Uri.SchemeDelimiter;
```
0x90d, line 36: Can it be made const (if not static readonly) and moved out of method?

0x90d (changes requested):

VDF.GUI/Utils/PickerDialogUtils.cs
0x90d: I only meant line 23, sorry

VDF.GUI/Utils/PickerDialogUtils.cs
jlesage: Since the scheme is determined from the `uriPath` (this way we don't have to do assumption about it), I think we can't ?

0x90d: You're right sorry, I read it as `Uri.Scheme` for some reason

VDF.GUI/Utils/PickerDialogUtils.cs
jlesage: Ok

VDF.GUI/Utils/PickerDialogUtils.cs
jlesage: Done with the latest update.

PR #416, opened 2023-05-08
Updated templates

David-Maisonave:
## Added prefix title
- Added a prefix title, so when user create the issue, it has an associated title prefix (Bugs, Feature-Request, Help), which makes it easier to see what type of issue it is before reviewing it.
- Also added a bug icon for the bug report template.
## Added associated labels
Added associated labels, so you don't have to set them manually.
## Added more details in the template body
- Template for Bug-Report did not have "Any" for OS, even though it was in Feature-Request. Added "Any" to bug report.
- Added minor improvements to template body.
## Help Template 
- Added help template for users posting questions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,7 @@
 ---
-name: 🐞 Bug report
+name: Bug report
 about: Create a bug report
-title: "🐞: "
+title: "[Bug]: "
 labels: "bug"
 assignees: "0x90d"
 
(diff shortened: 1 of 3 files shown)


0x90d:
Please change `Any` to `MacOS` and remove the bug icon.

David-Maisonave:
> Please change Any to MacOS and remove the bug icon.

Done.

By the way, I just figured out how to get YAML file to work for this, and the YML file is better because you can specify optional and required fields for the user to fill.
Take a look at the Bug and Feature request in the following link, which are using YML file.
https://github.com/David-Maisonave/MediaFileDuplicateFinder/issues/new/choose

Source code is [here](https://github.com/David-Maisonave/MediaFileDuplicateFinder/tree/main/.github/ISSUE_TEMPLATE).

If you like the YML version, I can change this pull request to use YML instead of MD files.

0x90d:
I don't have any interest into something like that. Nothing of this makes "my work" easier and this project don't get that many issues that it would be necessary.

Issue #306, opened 2022-03-04
Mega-groups


whacklezz:
**Background**
I have found that running the scan repeatedly with decreasing % similarity gives me good results, by gradually chipping away very good matches and flagging false positives. When coming down into the Mid/High 80s % similarity, I can still snag 30-40 groups of true duplicates with only 2 members in each group for the most part. However, there is always one "mega-group" with tens of thousands of members. This is due to how groups are merged if a sample is similar to any one member of the existing group, so it has a tendency of growing exponentially. It's for all intents and purposes impossible to visually compare these `n^2`, so I always just end up minimizing the group and waiting until it has processed all the thumbnails (which takes a good while eventually due to the high number).

**Describe the solution you'd like**
We need a way to break up these colossal groups. Surely many of the members of the group are only similar to one or a few others. I have a feeling that the union of all these sets is barely overlapping, but it ends up daisy-chaining everything into an all-consuming black hole. I guess it could also be one or a few "super-matchers" which somehow ends up reporting a high similarity to an overwhelming lot of others?

It could actually be interesting to see all the elements on a weighted graph. There probably exists theory to break up, cluster, or untangle such sets.

Maybe it could be possible to get the listed % similarity to change based on the element selected, and let us sort the list? Then at least we could expect any true matches to be close to the top.

Or run a separate scan on the group elements where they are broken up into distinct groups instead of merged into one. I guess both (x, y) and (y, x) would appear as groups though. It would of course mean potentially tens of thousands of groups instead of one group of tens of thousands.

Maybe we just need further options for filtering similarity to reduce the group sizes, like e.g. not allowing files with too different aspect ratios (landscape/portrait) to be considered similar.

Idk, I'm just spitballing here, but I think it's an issue that eventually needs addressing. I feel like there are a lot of dupes hiding somewhere inside that monster set :p

floydcg:
I've noticed the same thing, and actually if you take the time to look through them, there are often matches. Sometimes several of them. Perhaps also some flag or something maybe that if the group is larger than (x) # of duplicates, Ignore that dupe group?

0x90d:
> maybe that if the group is larger than (x) # of duplicates, Ignore that dupe group?

VDF doesn't know if duplicates are false positives or positive positives. There could be a group larger than X which only contains real duplicates. One way to reduce the amount of false positives is to increase the thumbnail size. Also this suggestion #300 could help to reduce the amount of false positives even further.

[...]

PR #438, opened 2023-07-14
New feature: Blacklisting the grouping of confirmed non-matching file pairs

whacklezz:
**Background**
When dealing with large groups, the likelihood of an item being added to the group grows with the group size. Similarity between false positives tends to create daisy-chains that eventually merges into super-groups.

If we explicitly declare items as "not matching"/"blacklisted", we can break these daisy chains and thus reduce them to smaller groups that are easier to manage. A simple daisy-chain of similar items can looks like this:

A-B-C-D-E-F-G-H

If we declare item D and E not matching, they are forbidden from grouping together and we end up with this group configuration instead:

A-B-C-D
E-F-G-H

**Summary**
Added scan-time blacklisting of the grouping of user-confirmed not-matching file pairs.
The files need to be scanned/rescanned in order to rearrange the already displayed groups after adding to the blacklist,
Tried to add inline comments to explain what each part of the code does.

**Adding to the blacklist**
Added a json file that keeps a dictionary of hashsets, one for each file that has a blacklisted pairing. It creates a two-way set, so you can look up A see that it blacklists B, and you can look up B and see that it blacklists A.

Initially the file does not exist. The entries are created from the right click menu of the scan result. When you right click an item, you find new options in the Selected Group submenu: 
1) To blacklist the selected item from matching with the each of the other items in the group
2) To blacklist each item in the group from matching with each of the other items in the group

When the option is selected, it adds entries to the dictionary file into the corresponding hashsets.

**Using the blacklist**
When scanning/rescanning the files, the duplicate check will run as usual. The difference is that it also reads the blacklist dictionary, and uses some rules when it compares items for similarity.

If it encounters an item with a blacklist, it checks the elements on the list to evaluate whether or not to group the candidate items. I have tried to comment on the logic to make it understandable.

**Blacklisting logic**
When comparing A and B and they are found to be similar:
- If neither is a member of an existing group:
 We create a new group unless they blacklist each other.
- If A is a member of an existing group and B is not:
 We can't add B to the group if there is a blacklist entry between B and one of A's group members (and similar the other way around). 
 Note: We could potentially move A from its existing group into a new group with B, but that means there can be a disconnect inside A's group where an item is not similar to any remaining items. Not implemented.
- If A and B are members of different existing groups: 
 If A blacklists one of B's members AND B blacklists one of A's members, we cannot merge the groups
 If either A blacklists one of B's members OR B blacklists one of A's members, we still cannot merge the groups (again, could move item from one group to another, e.g. because of higher similarity, but not implemented due to complexity).
If neither A nor B blacklists a group member of the other group, but a member of group A blacklists a member of group B, we cannot merge the groups. (Again, could move item from one group to another, but not implemented due to complexity)

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -404,6 +404,7 @@ namespace VDF.Core {
 		}
 
 		void ScanForDuplicates() {
+			Dictionary<string, HashSet<string>> BlacklistDictionary = new();
 			Dictionary<string, DuplicateItem>? duplicateDict = new();
 
 			//Exclude existing database entries which not met current scan settings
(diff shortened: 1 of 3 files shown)


whacklezz:
Addresses #306

PR #474, opened 2023-12-19
add Highlight Best Results, Non-Best Results, Duration, Size options to Appearance settings

sean-adams:
The functionality to highlight the best & non-best results can be helpful but when there are multiple duplicates, each of which scores best on various attributes, it can get a bit confusing / visually taxing, especially if there are certain attributes the user doesn't particularly care about highlighting.

This PR adds four new settings:

"Highlight Best Results" and "Highlight Non-Best Results", both with a default value of true (keep existing behavior to show best matches in green / non-best matches in red)
- Disabling "Highlight Best Results" means best matches will be shown in standard theme text color.
- Disabling "Highlight Non-Best Results" means non-best matches will be shown in standard theme text color.
- Disabling both means all match information will be shown in standard theme text color.

"Highlight Duration" and "Highlight Size", both with a default value of true (keep existing behavior to show best / non-best matches on Duration and Size (file size) attributes)
- disabling either of these will no longer show best Duration or Size in highlighted colors
- if this works, I can expand this PR to provide options for all "Best" attributes

Built & tested on osx-arm64.

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -119,10 +119,8 @@ namespace VDF.Core {
 			SearchTimer.Stop();
 			ElapsedTimer.Stop();
 			Logger.Instance.Info($"Finished scanning for duplicates in {SearchTimer.Elapsed}");
-			if (Settings.HighlightBestResults) {
-				Logger.Instance.Info("Highlighting best results...");
-				HighlightBestMatches();
-			}
+			Logger.Instance.Info("Highlighting best results...");
+			HighlightBestMatches();
 			ScanDone?.Invoke(this, new EventArgs());
 			Logger.Instance.Info("Scan done.");
 			DatabaseUtils.SaveDatabase();
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -119,8 +119,10 @@ namespace VDF.Core {
 			SearchTimer.Stop();
 			ElapsedTimer.Stop();
 			Logger.Instance.Info($"Finished scanning for duplicates in {SearchTimer.Elapsed}");
-			Logger.Instance.Info("Highlighting best results...");
-			HighlightBestMatches();
+			if (Settings.HighlightBestResults) {
+				Logger.Instance.Info("Highlighting best results...");
+				HighlightBestMatches();
+			}
 			ScanDone?.Invoke(this, new EventArgs());
 			Logger.Instance.Info("Scan done.");
 			DatabaseUtils.SaveDatabase();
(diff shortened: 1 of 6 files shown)


0x90d:
Why would it be useful to split up the "highlight (non)-best results"? If either of the two is activated both would still be highlighted in some way due to different colors. If color one has green/red the other is still default which still indicates what is considered best and what not. Unless Im missing something i would combine these two into one setting.

For size and duration I believe its better to perform the cheap boolean check first and then cast the object to string. As when user kept default setting it doesn't allocate a string.

hcucumber:
here is  a bug when i run VDF 3.0.x, the URL don't exist. My OS is ubuntu 2204.
Unhandled exception. System.InvalidOperationException: Default font family name can't be null or empty.
   at Avalonia.Media.FontManager..ctor(IFontManagerImpl platformImpl)
   at Avalonia.Media.FontManager.get_Current()
   at Avalonia.Media.Typeface.get_GlyphTypeface()
   at Avalonia.Rendering.Composition.Compositor.get_DiagnosticTextRenderer()
   at Avalonia.Rendering.Composition.Compositor.CreateCompositionTarget(Func`1 surfaces)
   at Avalonia.Rendering.Composition.CompositingRenderer..ctor(IRenderRoot root, Compositor compositor, Func`1 surfaces)
   at Avalonia.Controls.TopLevel..ctor(ITopLevelImpl impl, IAvaloniaDependencyResolver dependencyResolver)
   at Avalonia.Controls.WindowBase..ctor(IWindowBaseImpl impl, IAvaloniaDependencyResolver dependencyResolver)
   at Avalonia.Controls.WindowBase..ctor(IWindowBaseImpl impl)
   at Avalonia.Controls.Window..ctor(IWindowImpl impl)
   at Avalonia.Controls.Window..ctor()
   at VDF.GUI.Views.MainWindow..ctor() in /home/runner/work/videoduplicatefinder/videoduplicatefinder/VDF.GUI/Views/MainWindow.xaml.cs:line 37
   at VDF.GUI.App.OnFrameworkInitializationCompleted() in /home/runner/work/videoduplicatefinder/videoduplicatefinder/VDF.GUI/App.xaml.cs:line 30
   at Avalonia.AppBuilder.SetupUnsafe()
   at Avalonia.AppBuilder.Setup()
   at Avalonia.AppBuilder.SetupWithLifetime(IApplicationLifetime lifetime)
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder builder, String[] args, ShutdownMode shutdownMode)
   at VDF.GUI.Program.Main(String[] args) in /home/runner/work/videoduplicatefinder/videoduplicatefinder/VDF.GUI/Program.cs:line 30

PR #504, opened 2024-05-04
feat: add worst image size field for photo items

vin-rmdn:
Preface: I've been using this software for years now, thank you for making such a great software for sorting bulk-sized photos and videos - this has helped me in sorting my photos better without managed software (e.g., Mac Photos).

I have a problem: I have many duplicates of photos, and some of them contains thumbnails and resized images (thanks, Mac Photos!). I managed to save all these images, but handling 6k images manually have been a year-long hassle for me. With this I introduce IsWorstFrameSize, which picks only one image item that is the worst image quality, such that:
- it only picks the worst image
- it only picks one
- it does not pick anything if there are many identical photos in a group.

I suppose we can fix the IsBestFrameSize to only choose the one best frame size (and not choose all of them), but for this MR I'm adding a worst image that we can pick and delete images/videos with.

Open for discussion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -259,3 +259,6 @@ paket-files/
 # Python Tools for Visual Studio (PTVS)
 __pycache__/
 *.pyc
+
+# Project output
+outputFolder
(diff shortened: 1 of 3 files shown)


vin-rmdn:
Example query on using `item.IsWorstFrameSize` in Expression Builder as follows.
image

PR #517, opened 2024-06-19
Update ImageSharp to 3.1.4 (Fixes #516)

### LINKED ISSUE #516: [Bug]: Update ImageSharp to support webp
## Environment
 - OS: Linux

## Describe the bug
ImageSharp 2+ supports webp but cd0f2959f218f57fed82f57a0d57c467440c501c reverted it to 1.0.4.

I made a contribution a long time ago adding support for webp but around the same time ImageSharp was downgraded so it no longer works and all webps are imported as black squares. About since then (just under 2 years ago now) I've reverted this commit and everything still works perfectly, making me wonder why this was ever reverted in the first place.

So, given that VDF works perfectly with newer versions of ImageSharp and is even improved by it, can it please be updated?

kattjevfel:
This reverts commit cd0f2959f218f57fed82f57a0d57c467440c501c.

From my testing over the last 2 years of reverting this commit, everything works perfect with it, and it allows for webp's to be scanned properly.

### DIFF
--- a/VDF.Core/Utils/GrayBytesUtils.cs
+++ b/VDF.Core/Utils/GrayBytesUtils.cs
@@ -41,12 +41,15 @@ namespace VDF.Core.Utils {
 		public static unsafe byte[]? GetGrayScaleValues(Image original, double darkProcent = 80) {
 			// Lock the bitmap's bits.  
 			using var tempImage = original.CloneAs<Bgra32>();
-			tempImage.TryGetSinglePixelSpan(out var pixelSpan);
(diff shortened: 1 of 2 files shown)


0x90d:
Thanks for the PR. Can you try and update it to the latest ImageSharp which is already 3.x and see if that works too? If we bump the version, I prefer the latest if possible.

kattjevfel:
To my surprise it actually works perfectly with 3.1.4, which should be the latest as of right now. I tested with a fresh dataset and it included them and generated thumbnails perfectly.

PR #534, opened 2024-09-16
Better support for hardlink detection

KaibutsuX:
The previous approach for linux was scanning the root of the given file: /home/user/files/video_file1.mp4 => scanned: => / for all files

With a 30 second timeout generally would never complete before timing out. And if it did complete, it was scanning the entire filesystem for every single possible duplicate.

Now the scan will occur multiple times but only at the base of each included directory:
IncludeDir: /home/user1/files
IncludeDir: /home/user2/videos/
Scans => /home/user1/files/ and /home/user2/videos/

Ideally, I think the hard links approach for linux should simply collect all inode values at the start of file enumeration (st_ino) and use those for comparison (without even invoking ffmpeg or any other heuristics) since the inode number is the definition of hard-links (for linux), but that approach requires modifying the FileEntry struct which involves protobuf attributes and I'm not really familiar with that.

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -438,6 +438,9 @@ namespace VDF.Core {
 
 			//Exclude existing database entries which not met current scan settings
 			List<FileEntry> ScanList = new();
+			var includedPathsArr = new String[Settings.IncludeList.Count];
+			Settings.IncludeList.CopyTo(includedPathsArr);
+			var includedPaths = includedPathsArr.ToList();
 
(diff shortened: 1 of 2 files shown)


0x90d:
I'm not too familiar with linux but isn't your approach failing to detect a hard link if the link is outside of the scan directories? If the problem is that it takes longer than 30 seconds on a system with too many files then I think it would be better to make the timeout customizable.

0x90d:
The way to look for hardlinks has been improved in one of the latest commits.

PR #535, opened 2024-09-16
Track inodes and use them for hard link comparisons on linux

KaibutsuX:
This implements inode comparisons for accurate hard link detection in linux.

I'm not sure what affects the ProtoMember changes have by adding a new field to the FileEntry struct.

### DIFF
--- a/VDF.Core/FileEntry.cs
+++ b/VDF.Core/FileEntry.cs
@@ -38,6 +38,14 @@ namespace VDF.Core {
 			DateCreated = fileInfo.CreationTimeUtc;
 			DateModified = fileInfo.LastWriteTimeUtc;
 			FileSize = fileInfo.Length;
+			if (CoreUtils.IsWindows) {
+				Inode = 0;
+			} else {
+				var ok = Mono.Unix.Native.Syscall.stat(_Path, out var stat);
+				if (ok == 0) {
+					Inode = stat.st_ino;
+				}
+			}
 		}
 
 		[ProtoMember(1)]
@@ -65,6 +73,8 @@ namespace VDF.Core {
 		public DateTime DateModified;
 		[ProtoMember(8)]
 		public long FileSize;
+		[ProtoMember(9)]
+		public ulong Inode;
 
 		[ProtoIgnore]
 		internal bool invalid = true;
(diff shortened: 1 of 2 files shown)


KaibutsuX:
@0x90d This PR is preferable to my other hard link pr because it doesn't rely on any external utils like the find command.

0x90d:
Same as in your other PR. This seem to only exclude files where the 'real' file and the 'link' file are both part of the scans but that is not how excluding hard link feature work in VDF. See the windows implementation. It will exclude files which are hard links even if the 'real' file is outside of the scan directories.

0x90d:
> I'm not sure I understand the purpose of the setting then.

The purpose is that any file which is a hard link is excluded. It doesn't matter where it links to.

Thorongil80:
> > I'm not sure I understand the purpose of the setting then.
> 
> The purpose is that any file which is a hard link is excluded. It doesn't matter where it links to.

This is meaningless, I do not know how a hard link is implemented in Windows of all sorts. But on Linux, EVERY file in a file system is a hard link to an inode.

A usual file has a 1:1 ratio of "file names in the file system" (I call them "references" from now on) to its data.
A file that is referenced multiple times, has a n:1 ratio of references to its data.

NONE of the references is in any way special. There is not an "original file" and additional "hard links".

If you "exclude hard links" on Linux, you exclude the whole file. Because the references are uniform, none of them is "special".

If you want to only scan the file ONCE, you need to do so in the scanned directories. Check which references point to the same data and only KEEP ONE. 

Even though "hard links" are created using "ln <blabla> <link>", this does NOT create an "hard link". It looks up the target of <blabla> and links <link> to it. <blabla> and <link> are then the same THINGS, references pointing to the same data. Neither is <link> a hard link to <blabla> nor is <blabla> a hard link to <link>.

0x90d:
The way to look for hardlinks has been improved in one of the latest commits.

KaibutsuX:
Tested newest build on master. Hard link detection is still scanning OUTSIDE of the the user's selected scan directories (which slows down the search considerably) and ignores the user's request to scan only the designated directories.

Additionally, it doesn't appear to actually work. It still detects hard linked files (files pointing to the exact same inode) as duplicate files.

image

PR #549, opened 2024-11-29
add instructions about signing the binary on macOS

adamhl8:
I had trouble getting the program to run on macOS. Eventually figured out that macOS kills the process because its not signed. I added instructions to the readme for this.


adamhl8:
@0x90d One of your more recent commits (mistakenly?) removed these changes.

PR #562, opened 2024-12-31
make compare tolerant against slight variations in thumbnails

### LINKED ISSUE #560: [Bug]: on video compare exits too soon if a single thumbnail differs too much
## Environment
 - OS: Windows

## Describe the bug
I have two videos of a talk at GPN. They seem to originate from different sources. I have to lower the threshold down to 80% percent until it is detected as a duplicate. Then it shows a similarity of 97%. I am comparing 50 thumbnails.
I did debug the compare algorithm and detected that it is checking each thumbnail against the threshold value individually.
It happens that the video had a significant delta in one of the thumbnails, so the full video was declared as "no match".
It is only detected as duplicate if the threshold value can tolerate the largest variation of all thumbnails.

To me this looks like a bug in the algorithm. It should be tolerant against slight variations of single thumbnails if most of the other thumbnails are a close match.

stephankn:
When comparing individual thumbnails allow variations unless they exceed a value where it can no longer fit into the acceptable threshold.

fixes #560

### DIFF
--- a/VDF.Core/ScanEngine.cs
+++ b/VDF.Core/ScanEngine.cs
@@ -421,19 +421,18 @@ namespace VDF.Core {
 				return difference <= differenceLimit;
 			}
 
-			float diff, diffSum = 0;
+			differenceLimit *= positionList.Count;
+			float diffSum = 0;
 			for (int j = 0; j < positionList.Count; j++) {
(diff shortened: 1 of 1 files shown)


0x90d:
Thank you 👍

PR #566, opened 2025-01-07
add instructions about signing the binary on macOS

adamhl8:
This is exactly the same as my previous PR: #549 

The changes were merged and then reverted (by mistake?) in another commit.


0x90d:
Sorry about that

PR #689, opened 2026-03-11
Fix explorer.exe process leak when opening file in folder on Windows

### LINKED ISSUE #688: [Bug]: explorer.exe process leak when opening file in folder on Windows
## Environment
 - OS: Windows

## Describe the bug
Each call to "Open in Folder" spawns a new `explorer.exe` process (parented to `svchost.exe`) that never terminates, even after the user closes the Explorer and VDF window. These processes accumulate over time, consuming system resources.

## Steps to reproduce
1. Open VDF and load or run a scan with results
2. Right-click a duplicate item → "Open in Folder"
3. Close the Explorer window that opens
4. Repeat steps 2–3 several times
5. Open Task Manager → Details tab → sort by name
6. Observe multiple `explorer.exe` processes accumulating, none of which terminate

## Snapshots
https://github.com/user-attachments/assets/6c0af499-3463-433b-b64f-773a0ca1ca51

Damus765:
Fixes #688 

## Problem

The call to `Process.Start` creates an new `explorer.exe` process that never exits afterward.
https://github.com/0x90d/videoduplicatefinder/blob/47d78f681d555bd63ac5be684d75b4a2d64c5a86/VDF.GUI/ViewModels/MainWindowVM.cs#L1075

## Solution

I initially tried to set `UseShellExecute` to `false`. But it didn't solve the issue.

So I decided to replace the call to `Process.Start` with the Windows Shell API [`SHOpenFolderAndSelectItems`](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shopenfolderandselectitems), which communicates directly with the Windows Shell without spawning a new `explorer.exe` process ourselves.

Other platforms remain unchanged.

## Additional notes

- The folder appears to open noticeably faster than with the previous `Process.Start` approach.
- When the target file no longer exists (e.g. deleted externally), a clear error dialog is shown instead of Explorer silently opening the home/quick access folder.

## How to test

1. Run VDF and load or perform a scan with results
2. Right-click a duplicate → "Open in Folder" several times
3. Check Task Manager → Details: no new `explorer.exe` processes should appear

### DIFF
--- a/VDF.GUI/Utils/ShellUtils.cs
+++ b/VDF.GUI/Utils/ShellUtils.cs
@@ -33,7 +33,7 @@ namespace VDF.GUI.Utils {
 			}
 			finally {
 				if (pidl != IntPtr.Zero)
-					CoTaskMemFree(pidl);
+					Marshal.FreeCoTaskMem(pidl);
 			}
 		}
 
@@ -44,8 +44,5 @@ namespace VDF.GUI.Utils {
 		[LibraryImport("shell32.dll")]
 		private static partial int SHOpenFolderAndSelectItems(
 			IntPtr pidlFolder, uint cidl, [In] IntPtr[]? apidl, uint dwFlags);
-
-		[LibraryImport("ole32.dll")]
-		private static partial void CoTaskMemFree(IntPtr pv);
 	}
 }


0x90d:
Since nothing has happened in this PR since 2 weeks I've implemented your approach. Thanks for the research and work you did.

PR #703, opened 2026-03-28
Fix extraction of ffmpeg lib files on Linux

AngeredBacterium:
Second commit from https://github.com/0x90d/videoduplicatefinder/pull/684

There's discussion on whether this is necessary in the original PR.


0x90d:
Since its just a small bug I've merged and will fix it myself, thanks!

PR #717, opened 2026-04-19
Fix native FFmpeg crashes by allocating padded swscale destination frames

marcmy:
## Summary

Fixes native FFmpeg crashes in the in-process thumbnail/scaling path by replacing the tightly packed destination buffer in `VideoFrameConverter` with a properly padded/aligned destination `AVFrame`.

## Root cause

The native path previously allocated the swscale destination image with:

- `av_image_get_buffer_size(..., align=1)`
- `av_image_fill_arrays(..., align=1)`

Under WinDbg + page heap, the crash was reproducible as an invalid write inside `swscale-9.dll` during scaling. The fault occurred in the destination write path, which strongly pointed to the destination buffer being too tightly packed for swscale's SIMD writes.

## Fix

`VideoFrameConverter` now:

- allocates a real destination `AVFrame`
- sets `format`, `width`, and `height`
- calls `av_frame_get_buffer(..., 64)` to get padded/aligned frame storage
- calls `av_frame_make_writable()` before each conversion
- passes that padded frame to `sws_scale`

## Validation

Before the fix:

- crash reproduced with native FFmpeg binding enabled
- WinDbg reported `INVALID_POINTER_WRITE_AVRF_c0000005_swscale-9.dll`

After the fix:

- full scan completed successfully
- duplicate results displayed successfully
- thumbnail retrieval completed successfully
- no crash

### DIFF
--- a/VDF.Core/FFTools/FFmpegNative/VideoFrameConverter.cs
+++ b/VDF.Core/FFTools/FFmpegNative/VideoFrameConverter.cs
@@ -14,22 +14,18 @@
 // */
 //
 
-using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 using FFmpeg.AutoGen;
 
 namespace VDF.Core.FFTools.FFmpegNative {
 	sealed unsafe class VideoFrameConverter : IDisposable {
-		private readonly IntPtr _convertedFrameBufferPtr;
 		private readonly Size _destinationSize;
-		private readonly byte_ptrArray4 _dstData;
-		private readonly int_array4 _dstLinesize;
+		private readonly AVFrame* _pConvertedFrame;
 		private readonly SwsContext* _pConvertContext;
 
 		public enum ScaleQuality {
 			FastBilinear,
 			Bilinear,
-			Bicubic,   // empfohlen
+			Bicubic,
 			Lanczos,
 			Spline,
 			Area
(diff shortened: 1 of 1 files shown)


VDF.Core/FFTools/FFmpegNative/VideoFrameConverter.cs
```diff
@@ -69,50 +65,37 @@ public VideoFrameConverter(Size sourceSize,
...
-				ref _dstLinesize,
-				(byte*)_convertedFrameBufferPtr,
-				destinationPixelFormat,
-				destinationSize.Width,
-				destinationSize.Height,
-				1).ThrowExceptionIfError();
+			// Give swscale a real padded destination frame instead of a tightly packed align=1 buffer.
+			// The WinDbg trace showed an invalid write inside swscale-9.dll while writing the converted
+			// destination image, so we intentionally request aligned/padded frame buffers here.
+			ffmpeg.av_frame_get_buffer(_pConvertedFrame, 64).ThrowExceptionIfError();
```
0x90d, line 79: Thank you for bringing my attention to it. Passing `1` was incorrect but I think we should follow FFmpeg recommendation here and use 0 instead of 64:

```
 * @param align Required buffer size and data pointer alignment. If equal to 0,
 *              alignment will be chosen automatically for the current CPU.
 *              It is highly recommended to pass 0 here unless you know what
 *              you are doing.
 ```
[Source](https://github.com/FFmpeg/FFmpeg/blob/415b466d41ac81856abc76d7a9341132b0f668b0/libavutil/frame.h#L883-L887)

0x90d:
Thanks for your PR and the research. I've applied it with passing 0 instead of 64 following ffmpeg recommendation

marcmy:
Sorry, I missed this ping/notification somehow. Thanks for the review and for catching the align value, using 0 makes sense per FFmpeg’s recommendation. Glad this is fixed now.

PR #743, opened 2026-05-05
Fix: FFmpeg/FFprobe not found on macOS & Linux

viniciusmi00:
There is a bug in Video Duplicate Finder on macOS and Linux where the desktop app fails to detect the FFmpeg and FFprobe binaries, even when they are properly installed (see attached images from macOS).
I reproduced the issue on macOS Tahoe (both Intel and Apple Silicon) and on an Ubuntu virtual machine, where the error occurred consistently.

This PR fixes the detection issue. The solution was tested on:
	•	macOS Tahoe (Intel and Apple Silicon)
	•	Linux (Ubuntu VM)
	•	Windows 11
	
Images:
Captura de Tela 2026-05-04 às 09 32 24

Captura de Tela 2026-05-04 às 09 33 04

<br>

Additionally, I introduced two input fields that allow users to manually specify custom paths for FFmpeg and FFprobe (see image below). This helps users who installed the binaries in non-standard locations. Once a valid full path is provided, the application correctly recognizes the binaries.

Image:
Captura de Tela 2026-05-05 às 04 36 37

<br>

This change both resolves the detection bug and adds flexibility for custom configurations.

Tested and working on:
	•	macOS (Intel and Apple Silicon)
	•	Linux (Ubuntu)
	•	Windows 11
Not tested on:
	•	Web
	•	CLI
	•	Docker
	•	Other Linux distributions  
	
**Note 1:** I added UI text in English and Brazilian Portuguese (my native language) only. I avoided other languages to prevent incorrect translations.
If you'd like translations in additional languages, I can generate them using AI tools and include them in this PR.

**Note 2:** The last build I tested was from May 4, 2026.


0x90d:
Thanks for the PR, the Homebrew fallback is a real gap and worth fixing. A few questions and concerns before this can go in:

### Blocking

1. **Hardcoded `\\` separators break macOS/Linux.** In `VDF.Core/FFTools/FFToolsUtils.cs` the original used `Path.Combine(CoreUtils.CurrentFolder, "bin", toolExecutable)`. The new code uses `$"{CoreUtils.CurrentFolder}\\bin\\..."`. Backslash is Windows-only, on macOS/Linux that string is a single filename and `File.Exists` will never match. Can you restore `Path.Combine` and reintroduce the `toolExecutable` local?

2. **The custom path may not take effect at runtime.** `FfmpegEngine.FFmpegPath` (line 37) and `FFProbeEngine.FFprobePath` (line 24) are cached in **static constructors**. If those types are touched before `SettingsFile` writes `FFToolsUtils.CustomFFmpegPath` (e.g. via the FFmpeg downloader at `MainWindowVM_FfmpegDownloader.cs:95`), the cached value is stale and editing the field in Settings does nothing until restart. Could the setter re-resolve and reassign `FfmpegEngine.FFmpegPath` / `FFProbeEngine.FFprobePath`, or could the static caching be removed in favor of lazy lookup?

3. **The setting is dead in CLI / Web / Docker.** `Settings.CustomFFmpegPath/Path` was added to `VDF.Core/Settings.cs`, but only the GUI's `SettingsFile` setter pushes it into `FFToolsUtils`. `VDF.CLI/Commands/ScanRunner.cs` deserializes `Settings` directly and never copies the new fields. The Web Settings page binds inputs that look like they work but the value is never read by `FFToolsUtils.GetPath`. Could you wire CLI/Web startup to copy these into the `FFToolsUtils` statics (or remove the statics, see point 4)?

### Design

4. **Three copies of the same value.** It now lives in `Settings`, `SettingsFile`, and as static fields on `FFToolsUtils`. Could `GetPath` read directly from `Settings.Instance.CustomFFmpegPath` (one source of truth) instead of mirroring through statics?

5. **No `Browse…` file picker.** Avalonia's file dialog avoids paste/typo errors. Could you add a Browse button next to each TextBox?

6. **No validation feedback.** A bad path silently falls through to PATH lookup, so users with a typo will think the override is broken. Could you log/warn when a configured custom path is rejected, or surface a "File not found" hint on lost focus?

7. **Label says "(Mac/Linux)" but the code applies on Windows too.** The custom-path branch runs before any OS check. Either drop the parenthetical from the label or guard by OS.

### Style / project conventions

8. `Settings.cs` - the new lines are indented with spaces but the file uses tabs (per `.editorconfig`).
9. `FFToolsUtils.cs` - the comment `// 4. FALLBACK MACOS: Se não encontrou no PATH…` is in Portuguese and references a numbering ("4.") that has no 1/2/3. Could you translate to English and drop the number?
10. The repeated `tool == FFTool.FFmpeg ? ffMpegPlatformName : ffProbePlatformName` ternary appears four times where the original had one `toolExecutable` local, could that local be reintroduced?
11. Strings were added only to `en.json` and `pt.json`. Could you also add them to `de`, `es`, `fr`, `zh-Hans` (copying the English values is fine as a fallback) so those locales don't show the raw key?
12. Missing trailing newline on `.gitignore`.
13. `Settings.razor` - the existing "Custom FFmpeg arguments" block was reindented from spaces to tabs, which adds noise to the diff. Could you revert the indentation-only changes on the pre-existing label?

### Tests

14. Could you add a unit test in `VDF.Core.Tests` covering the `GetPath` precedence (custom → local `bin` → app dir → PATH → macOS fallback)? Without it, the next refactor will silently regress.

The Homebrew fallback alone is a clean, mergeable fix on its own, happy to land that piece quickly if you'd prefer to split it from the custom-path UI.

viniciusmi00:
@0x90d I got sick. Flu or Covid. Give me a few days and I'll answer that and fulfill your requests.

0x90d:
No problem. Get well soon.

0x90d:
Thanks again for spotting and digging into this, @viniciusmi00 - the FFmpeg/FFprobe detection failure on macOS and Linux was a real bug and your report (plus the macOS screenshots) made the root cause clear.

Since you've been unwell, I didn't want to leave this hanging or pressure you to do a round of review changes while you're recovering, so I went ahead and landed the detection fix directly in 35d048f.

PR #785, opened 2026-06-08
Add Video Compare integration (group actions + custom launch arguments)

HamzaETTH:
## Summary

- Allow dragging one or multiple videos from the duplicate grid to external programs
- Added **Video Compare** button in each duplicate group header action row
- Added **Video Compare** settings section (executable path, custom arguments, link to the [video-compare project](https://github.com/pixop/video-compare))
- Custom arguments support shell-style quoting (single/double quotes, escaped quotes)

### DIFF
--- a/VDF.GUI/Data/SettingsFile.cs
+++ b/VDF.GUI/Data/SettingsFile.cs
@@ -56,6 +56,20 @@ namespace VDF.GUI.Data {
 			public string OpenMultiple { get; set; } = string.Empty;
 		}
 
+		string _VideoCompareExecutablePath = string.Empty;
+		[JsonPropertyName("VideoCompareExecutablePath")]
+		public string VideoCompareExecutablePath {
+			get => _VideoCompareExecutablePath;
+			set => this.RaiseAndSetIfChanged(ref _VideoCompareExecutablePath, value);
+		}
+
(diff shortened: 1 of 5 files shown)


0x90d:
I don't think it would be a good idea to add buttons specifically for a particular program. Then someone else would come along and want their program supported as well, and so on. That would be better suited for a) plugins or b) a fork. What we could do instead is add a `Open this group in an external program` feature, and users could then choose any external program they want.

But we already have that, with `Custom Commands` under `Processing`

HamzaETTH:
I would normally agree, but this feels like the one missing piece that would make VDL feel complete.

video-compare is extremely lightweight and minimal. It doesn't really introduce any significant UI beyond showing the two videos side by side, so it doesn't feel out of place within VDL. Having it integrated directly would also save a lot of time and friction compared to launching a separate tool.

That said, I understand your concern about scope and maintenance. Either way, I wanted to make the case for it.

0x90d:
VDF already has its own video comparer that displays both videos side by side and offers features such as frame-by-frame stepping. It’s not as if VDF has absolutely no video comparison features. But yes, it’s clearly not as feature-rich as the video comparer tool. However, I think it fits better in the "external tool" category rather than having its own section in this app.

PR #797, opened 2026-06-13
feat(cli): add 'db export' command for headless database JSON export

awernick:
Adds `vdf-cli db export --output path.json` for headless database JSON export. Uses the existing `ScanEngine.ExportDataBaseToJson()` public API. Useful for scripting/automation pipelines that need the full database dump (fingerprints, media info, pHashes) without the GUI.

### DIFF
--- a/VDF.CLI.Tests/Commands/DatabaseCommandTests.cs
+++ b/VDF.CLI.Tests/Commands/DatabaseCommandTests.cs
@@ -33,7 +33,7 @@ public class DatabaseCommandTests {
 		var cmd = DatabaseCommand.Build();
 		var root = new RootCommand { cmd };
 		var result = root.Parse("db export -o /tmp/test.json");
-		Assert.Equal(0, result.Errors.Count);
+		Assert.Empty(result.Errors);
 	}
 
 	[Fact]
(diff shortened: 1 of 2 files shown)


0x90d:
Thanks for this, clean, well-scoped, and follows the existing `db` subcommand pattern. Builds and tests pass. A few things before merge:

1. **`--output` should be `Required = true`** instead of optional `FileInfo?` + a manual null-check. As written, a missing `--output` prints an error but exits **0**, which a script can't detect. `Required = true` gives a proper parse error, non-zero exit, and better `--help` for free.

2. **Failure paths return exit 0** (empty DB, export failed). For an automation command this should be non-zero, return `Task<int>` and `return 1` on those paths. Heads up: `db clean`/`db clear` have the same gap, so consider fixing exit codes across all three consistently (happy to take that as a follow-up if you'd rather keep this PR minimal).

3. **xUnit2013 warnings:** `Assert.Equal(0, result.Errors.Count)` → `Assert.Empty(result.Errors)` in both parse tests.

Minor: the tests only cover registration/parsing. An end-to-end test writing a small DB to a temp JSON would be nice. And `IncludeFields = true` is a no-op here (the core export only reads `WriteIndented`), but it mirrors the GUI so no need to change it.

awernick:
@0x90d thanks for the review! Latest commits address your 3 comments. For 2, I kept changes to a minimum but happy to open that follow up PR to clean up `db clean / db clear`

PR #813, opened 2026-06-25
Enabled Web-UI to export Selection Status as well

marioli:
Without this patch, the selection is not reflected in the export. This patch extends the CSV export by the "selected" attribute.

### DIFF
--- a/VDF.Web/Components/Pages/Results.razor
+++ b/VDF.Web/Components/Pages/Results.razor
@@ -755,7 +755,12 @@ else {
 
     // === CSV Export ===
     async Task ExportCsv() {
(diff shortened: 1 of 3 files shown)


0x90d:
Thanks for this!  One small thing before merge: the new `Selected` column encodes the boolean as `checked/unchecked`, but the GUI's equivalent column (Checked) emits `True/False` via `bool.ToString()` ([MainWindowVM.cs:827](https://github.com/0x90d/videoduplicatefinder/blob/master/VDF.GUI/ViewModels/MainWindowVM.cs#L827)). Could you switch the Web export to True/False to match? That way anyone parsing both CSVs gets a consistent encoding of the same field. The column-name difference (Selected vs Checked) is already noted in your comments and is fine to keep.

PR #883, opened 2026-08-05
Refactor progress reporting in ScanEngine

### LINKED ISSUE #868: [Enhancement]: Auto save progress scheduler
### Operating System

Windows

### VDF variant

Desktop (GUI)

### Describe the desired feature

Idea: instead of asking the user to manually pause and resume a scan to save the current progress, it might be a nice feature to add a scheduler option to do this.

### What issue does it solve?

If you scan a bigger library and your system or program crashes, you lose every data since you pressed pause/resume last time. This feature would reduce the loss to the timeframe of the last saved version.

### Snapshots

_No response_

AngeredBacterium:
This PR deduplicates and merges much of the progress reporting logic done in `ScanEngine`, which should resolve #868, then extracts it from `ScanEngine`.

The first commit is just a minor cleanup. The second is the refactoring. The third then extracts all possible logic from `ScanEngine` into a new type and replaces the struct used for snapshots with a record. It being a record updated using `x with {}` means only values that might change need to be applied by callers, and being an immutable reference type means all the locks can go. Switching to a new type means the signature of the `ScanEngine.Progress` event changes, but all the fields are identical.

It's possible that after the extraction the update logic could be cleaner (removing some of the private fields in `ScanEngine` reapplied to updates would be nice) but I didn't feel up to tracking down all the flows and calls to verify that.

If you don't want the extraction part, or want it PR'd separately, just let me know and I'll drop it from this.

### DIFF
--- a/VDF.Core.Tests/AiPartialPhaseVisibilityTests.cs
+++ b/VDF.Core.Tests/AiPartialPhaseVisibilityTests.cs
@@ -59,7 +59,7 @@ public class AiPartialPhaseVisibilityTests : IDisposable {
 		var events = new ConcurrentQueue<ScanProgressChangedEventArgs>();
 
 		// Stand in for the visual gate that ran just before: it ends full over its own maximum.
-		engine.InitProgress(12);
+		engine.InitProgress(12, "");
 		for (int i = 0; i < 12; i++)
 			engine.IncrementProgress("clip.mp4");
 		engine.Progress += (_, e) => events.Enqueue(e);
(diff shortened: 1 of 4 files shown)


0x90d:
Thanks for this, and for the sharp root-cause analysis in #868!

Reviewed and tested: everything builds, all test suites pass, and the first two commits fix the checkpoint starvation exactly right. I'll take those as-is.

For the extraction commit (f1474dc) I'll take you up on your offer to split it into its own PR. The design looks good, but it changes VDF.Core's public event API and I'd rather review that separately. When resubmitting, please match house style: tabs instead of spaces (see `.editorconfig`), no trailing whitespace, block-scoped namespaces, no `private` modifiers.

So: just drop the third commit from this PR. I'll add "Fixed #868" when merging so the issue auto-closes, and add a regression test for the starvation afterwards.

AngeredBacterium:
Ok! Last commit dropped.

PR #887, opened 2026-08-08
Extract progress reporting logic from `ScanEngine`

AngeredBacterium:
Originally in #883, now with the style fixes applied.

Extracts all progress logic from `ScanEngine` into a new type and replaces the struct used for snapshots with a record. It being a record updated using `x with {}` means only values that might change need to be applied by callers, and being an immutable reference type means all the locks can go. Switching to a new type means the signature of the `ScanEngine.Progress` event changes, but all the fields/properties of the record are identical to the original struct.

It's possible that after the extraction the update logic could be cleaner (removing some of the private fields in `ScanEngine` reapplied to updates would be nice) but I didn't feel up to tracking down all the flows and calls to verify that.


0x90d:
Thanks, the extraction looks good and the full test suite passes on the branch. One behavioral regression to fix before merging:

The old code cleared the heartbeat snapshot when a scan ended (`isScanning` setter, `hasProgressSnapshot = false`). `ScanProgress` keeps its last snapshot forever, so on the next scan in the same session the heartbeat repaints the previous scan's final numbers while the new scan is still building its file list - that phase can run for minutes on large libraries before the first `InitProgress` fires. Resetting the snapshot to Default at scan end (or at scan start, before the heartbeat timer is armed) restores the old behavior.

Two minor points, neither a blocker:

- `InitProgress` used to include a Drives snapshot in its event; Reset now sends `Drives = null`, so the per-drive rows briefly vanish at each phase transition.
- House style also applies to the new file: no private modifiers and no underscore prefixes on fields (`ScanProgress.cs`, and the `_progress` field in `ScanEngine`).
Please rebase now that #886 is in, then this is good to go.

AngeredBacterium:
> The old code cleared the heartbeat snapshot when a scan ended

I've added a call to `Reset` with an empty transformation in the same place that `hasProgressSnapshot = false` used to be set, so the progress gets reset to default. `ScanProgress.Heartbeat` won't do anything [if the snapshot has reference equality with the default](https://github.com/AngeredBacterium/videoduplicatefinder/blob/1a20145bf1516a6ab30b3d924bc0ec6337bdcc40/VDF.Core/ScanProgress.cs#L81).

> * `InitProgress` used to include a Drives snapshot in its event;

No, it didn't. Before my previous change, [`InitProgress` never set `Drives` at all.](https://github.com/0x90d/videoduplicatefinder/blob/a7019a7ebd604f1da407c82644022ff34c17a27c/VDF.Core/ScanEngine.cs#L141-L148) Then, after my change, it only set it [if `driveProgressTracker` was not null](https://github.com/0x90d/videoduplicatefinder/blob/35360a891685c03e87267f29b2d3610ad9a1686c/VDF.Core/ScanEngine.cs#L285), but `driveProgressTracker` is set to null [in `InitProgress`](https://github.com/0x90d/videoduplicatefinder/blob/35360a891685c03e87267f29b2d3610ad9a1686c/VDF.Core/ScanEngine.cs#L136), which is unchanged from the original code.

> * House style also applies to the new file: no private modifiers and no underscore prefixes on fields (`ScanProgress.cs`, and the `_progress` field in `ScanEngine`).

This repository has more private fields prefixed with underscores than without, and your recent commits mix both styles.

If I try to set `.editorconfig` to enforce having no prefix, there are 390 errors. If I set it up to enforce having them, there are 149. Since there's no rule about this right now, my IDE defaults to adding them, since that's most common in this repo (and also the standard C# convention).

I would offer to fix that, but the 'official' recommended way to apply these types of changes seems to have a bug in my IDE, so I won't right now.

0x90d:
Thanks, the heartbeat fix works and you're right about Drives - `driveProgressTracker` is nulled before the phase-init push on master too, my mistake. Style changes appreciated; agreed the codebase is mixed on underscores.

One thing the fix commit introduces: `Reset(x => x)` also raises Progress with the Default snapshot, and since `isScanning = false` runs before `ScanDone` at every exit, the CLI now prints a `[  0%] 0/0  ETA ... line over the final [100%] N/N` (visible in terminals and as a garbage line in Docker logs). Master cleared the snapshot silently. Could you make the scan-end path a silent clear (no event), e.g. a Clear() that resets snapshot/lastUpdate without invoking Progress? A Core test asserting no event fires at scan end would be great. Then this is good to merge.

AngeredBacterium:
Ah, that was a dumb oversight. Thanks for catching it. Fixed that by just adding a slightly more explicit method that lacks the event invoking.

PR #896, opened 2026-08-20
Improve PHash performance & clean it up a bit

AngeredBacterium:
Noticed a few easy fixes/cleanups with computing PHashes. They all got it about 2x faster until I found out `TensorPrimitives` would work here, That got it 5-10x faster depending on the machine.

### Older Intel desktop CPU
```
| Method                | Mean      | Error     | StdDev    | Ratio | Allocated | Alloc Ratio |
|---------------------- |----------:|----------:|----------:|------:|----------:|------------:|
| ComputePHash_Original | 13.361 us | 0.0143 us | 0.0134 us |  1.00 |         - |          NA |
| ComputePHash          |  2.759 us | 0.0016 us | 0.0014 us |  0.21 |         - |          NA |
```

### Newer AMD laptop CPU
```
| Method                | Mean      | Error     | StdDev    | Ratio | Allocated | Alloc Ratio |
|---------------------- |----------:|----------:|----------:|------:|----------:|------------:|
| ComputePHash_Original | 21.030 us | 0.1218 us | 0.1079 us |  1.00 |         - |          NA |
| ComputePHash          |  2.153 us | 0.0089 us | 0.0083 us |  0.10 |         - |          NA |
```

### DIFF
--- a/VDF.Core.Tests/pHash/PerceptualHashTests.cs
+++ b/VDF.Core.Tests/pHash/PerceptualHashTests.cs
@@ -94,13 +94,13 @@ public class PerceptualHashTests {
 	}
 
 	/// <summary>
-	/// Bit-exact parity guard. The current implementation skips computing the 94% of DCT
-	/// coefficients that are never read, but its scalar accumulation order matches a
-	/// reference implementation that computes the full N×N DCT. This test pins that
-	/// equivalence — any future change that reorders adds (e.g. SIMD vectorization)
-	/// can produce slightly different floats, flip 1-2 bits at the median boundary,
-	/// and silently invalidate cached PHashes from older scans. If you intend to break
-	/// that compatibility, bump <c>DatabaseUtils.DbVersion</c> at the same time.
+	/// Tolerance guard. The current implementation skips computing the 94% of DCT
+	/// coefficients that are never read, and was changed to use SIMD when available
+	/// that, in over 99.9% of cases, should match a reference implementation that
+	/// computes the full N×N DCT. The outliers had a hamming difference of 2 or less.
+	/// This test maintains that tolerance. If you intend to break that compatibility
+	/// for any meaningful number of inputs, bump <c>DatabaseUtils.DbVersion</c> at the
+	/// same time.
 	/// </summary>
 	[Fact]
 	public void ComputePHash_MatchesFullDctReference() {
@@ -112,7 +112,9 @@ public class PerceptualHashTests {
 			ulong actual = PerceptualHash.ComputePHashFromGray32x32(gray);
 			ulong expected = ReferenceFullDct(gray);
 
-			Assert.Equal(expected, actual);
+			var distance = BitOperations.PopCount(expected ^ actual);
+
+			Assert.InRange(distance, 0, 2);
 		}
 
 		// Mirrors the previous N×N implementation: compute every DCT cell, then sweep the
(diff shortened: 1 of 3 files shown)


0x90d:
Thanks, this is a nice cleanup and I've verified the numbers on my side. On a 9950X3D, BenchmarkDotNet with your input:

| ISA forced | master | this PR |
|---|---|---|
| native (AVX-512) | 5.64 us | 2.13 us |
| AVX2 only | 6.50 us | 2.11 us |
| SSE only | 5.41 us | 3.36 us |
| intrinsics disabled | 5.49 us | 3.60 us |

So 2.6-3x here and no regression on the fallback paths. The CI failure is an unrelated flaky GUI test (#897 fixes it), and an AOT publish of the CLI from this branch is clean; pHash scan/compare works in the AOT binary.

One thing needs to change before I merge, though. The previous implementation deliberately kept the scalar accumulation order so hashes were bit-identical to the original full-DCT version, and therefore identical across machines and across scans. The comment block in `ComputePHashFromGray32x32` and the `ComputePHash_MatchesFullDctReference` test both document that contract. `TensorPrimitives.Dot` breaks it by design (SIMD lane accumulation, FMA where the CPU has it). I ran master and this branch over a 112k-frame corpus (random noise, blurred, low-contrast, dark fades, gradients, ffmpeg synthetic sources) with forced ISA levels:

- master vs this PR on the same machine: 11 of 112,000 frames differ, max Hamming distance 2
- this PR on AVX-512 vs AVX2 vs SSE vs scalar: 3-10 frames per 111,200 differ, max 2 bits
- master across the same ISA levels: 0 differences

The existing parity test only stays green because it samples pure random noise, which is the one input class that never drifts (0 of 20,000 in my corpus).

In practice 2 bits on 0.01% of frames is harmless: a pair would have to sit exactly on the threshold on exactly that frame to flip, and the quorum absorbs it. I don't want a `DbVersion` bump for this, that would force everyone through a full rescan for a rounding-level effect. But the guarantee should be dropped on purpose and documented, not lost silently. Concretely:

1. **Build the cos/alpha tables in double again** (`Math.Cos` / `Math.Sqrt`, then cast). The `MathF` variant adds its own drift (the argument rounding at `504*pi` is about 1e-4 rad) and buys nothing, since the tables are built once at static init. I benchmarked a double-table variant of your branch at 2.03 us, i.e. identical.
2. **Rewrite the comment block** in `ComputePHashFromGray32x32`. "Same scalar accumulation order as before. Floats are bit-identical to the previous implementation so cached PHashes from older scans remain valid" is no longer true. It should say that accumulation is SIMD-dependent and that hashes can differ from older scans or other machines by up to ~2 bits on a small fraction of frames.
3. **Turn `ComputePHash_MatchesFullDctReference` into a tolerance guard**: assert Hamming distance <= 2 against the reference over a few hundred inputs that include smooth and low-contrast frames, not just random noise, and update its doc comment accordingly.
4. **Add an explicit `<PackageReference Include="System.Numerics.Tensors" Version="9.0.0" />` to `VDF.Core.csproj`.** Right now `TensorPrimitives` only resolves transitively through `Microsoft.ML.OnnxRuntime.Managed`. A future ONNX bump could drop it (build break) or pull a newer `Tensors` with a different `Dot` kernel, which would shift the hashes again without anyone noticing.

Cosmetic, take it or leave it: `Alpha` is now eight identical values and could be a single constant.

With those four in, happy to merge.

AngeredBacterium:
All sound reasonable. I'm amused that I missed that `Alpha` was all the same values 😆

I applied that and points 1,2,4, but would you be alright handling the test data for point 3? Since you have a big pool of frames and can already find ones that changed.

PR #897, opened 2026-08-20
Fix 2 flaky tests

AngeredBacterium:
Fixes flaky tests that have failed PR builds:
`VDF.GUI.Tests.ResultsInteractionRulesTests.PathCopiedFlash_RapidReCopy_ExtendsInsteadOfCuttingShort`
https://github.com/0x90d/videoduplicatefinder/actions/runs/32352402341/job/96374209199#step:9:9
https://github.com/0x90d/videoduplicatefinder/actions/runs/31268921057/job/93131493575#step:9:9

`VDF.GUI.Tests.FolderCountingServiceTests.DuplicateRequests_AreDeduplicated`
https://github.com/0x90d/videoduplicatefinder/actions/runs/31756028544/job/94631827304#step:9:9

For `ResultsInteractionRulesTests.PathCopiedFlash_RapidReCopy_ExtendsInsteadOfCuttingShort`, I think it's due to thread/task contention. There's no guarantee that the first task gets to continue before the second, as there's no way to guarantee `Task.Delay` means "exactly this long". It has to mean "at least this long", since the thread pool can be busy. Making the delays artificial using a `FakeTimeProvider` solves this and also makes the test complete instantly.

For `FolderCountingServiceTests.DuplicateRequests_AreDeduplicated`, I think there was one real issue, one unlikely issue, and one fix just for simplicity/correctness.

First the correctness change. I don't think creating 200 files was necessary. The logic to start the first counting task, mark that path as being active, then try to start the second task all looks like it's synchronous on the same thread with nothing else inbetween (unless `Task.Run` can run the provided body synchronously in some cases. I don't know for sure that it *never* does, but I don't think it does). But that's just how `StartCounting` is implemented right now. So instead I changed it so it makes sure the progress callback is called after the 1 file, blocks continuing until the second counting tries & fails to start, then lets the first finish. This is hacky and relies on the `onProgress` callback being executed synchronously, but works well.

The "unlikely issue" is one I ran into when artificially trying to induce the flakiness. Stressing the thread pool and running this test hundreds of times in parallel could make the timeout run out and fail the test. I don't think any of what I've changed will properly fix that, maybe just mitigate it slightly as a side effect by not blocking the thread when waiting.

I think the "real issue" was that completion was reported via `onProgress` before the guard was actually removed. This would create a race condition where the third counting could try to start before the first had removed itself from the `active` dictionary. This is easily fixed by just calling the callback after the bookkeeping is done.

### DIFF
--- a/VDF.GUI.Tests/FolderCountingServiceTests.cs
+++ b/VDF.GUI.Tests/FolderCountingServiceTests.cs
@@ -71,17 +71,26 @@ namespace VDF.GUI.Tests {
 
 		[Fact]
 		public void DuplicateRequests_AreDeduplicated() {
-			// Enough files that the first walk is still running when the second starts.
-			for (int i = 0; i < 200; i++)
-				WriteFile($@"many\file{i}.mp4");
+			// Provide one file to ensure progress reporting happens before completion.
+			WriteFile(@"many\file.mp4");
 
-			var service = new FolderCountingService();
+			var service = new FolderCountingService(TimeSpan.Zero);
 			using var done = new System.Threading.ManualResetEventSlim();
-			bool first = service.StartCounting(root, p => { if (p.Completed) done.Set(); });
+			using var continueFirst = new System.Threading.ManualResetEventSlim();
+			bool first = service.StartCounting(root, p => {
+				if (!p.Completed) {
+					continueFirst.Wait();
+				}
+				else {
+					done.Set();
+				}
+			});
 			bool second = service.StartCounting(root, _ => { });
 
 			Assert.True(first);
 			Assert.False(second);
+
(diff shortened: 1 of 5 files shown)


AngeredBacterium:
~Ironically the PR build for this failed due to a 3rd flaky test: `VDF.GUI.Tests.FolderCountingServiceTests.Cancel_StopsWithoutCompletionReport` 🤣. I'll look into that.~

Edit: Nope, I was wrong. Forgot to run all tests after the last change and I'd actually missed something. That test does suffer from the same "unlikely issue". I think the best real fix for these would be to use something like `System.IO.Abstractions` and get rid of all the I/O and guesswork. But that's a bigger change.

PR #913, opened 2026-08-30
Fix logging test file path

AngeredBacterium:
This makes `CoreUtils.CurrentFolder` more reliably point to the application's directory by using `AppContext.BaseDirectory` instead of `Environment.ProcessPath`, then makes the logger tests no longer run in parallel and clean up before & after.

`Environment.ProcessPath` will point to the installed `dotnet` executable when running tests or if the user runs programs using `dotnet path/to/dll`.

I noticed this when one of the logger tests failed just once (another flaky test I guess) and found it had always been appending to a file inside my .NET installation directory, without ever cleaning it up. After fixing the path issue, I've made the logger tests no longer run in parallel and delete the log file before & after each test, just to make sure there are no future issues.

### DIFF
--- a/VDF.Core.Tests/Utils/LoggerTests.cs
+++ b/VDF.Core.Tests/Utils/LoggerTests.cs
@@ -18,10 +18,12 @@ using VDF.Core.Utils;
 
 namespace VDF.Core.Tests.Utils;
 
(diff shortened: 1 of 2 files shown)


0x90d:
♥️

